PR #37, opened 2026-08-04
feat: keep what the decoder knew, not just what it said

znat:
The audio has been on disk since the first version and the log has held the
transcript. Everything in between was thrown away one line after it arrived —
Parakeet returns a start, an end and a confidence for every token, and we took
`result.text` and dropped the rest.

This adds `trace.jsonl` beside the recordings: one JSON object per dictation,
appended, never rotated, joined to its clip by `wav`.

```json
{"at":"2026-08-04T07:59:55Z","wav":"parrotflow-2026-08-04T09-39-12.wav","source":"cli",
 "asr":{"text":"If no can we do this?","confidence":0.848,"duration":1.69,"processing":0.26,
        "words":[{"word":"If","start":0.16,"end":0.32,"confidence":0.582},
                 {"word":"this?","start":1.04,"end":1.28,"confidence":0.345}]},
 "vad":{"speech":1.53,"total":1.69,"segments":[[0.16,1.686]]},
 "stages":[{"name":"replacements","before":"…","after":"…","seconds":0.0014},
           {"name":"transform email","skipped":"app /mail|…/ did not match Ghostty"}],
 "final":"If no can we do this?"}
```

New beyond what the log had: the raw ASR text **before** substitutions, per-word
spans and confidences, the speech gate's segment boundaries rather than its
totals, per-stage cost in seconds, and every stage that ran — including the ones
that changed nothing, which the log conflates with "skipped".

## Why a second file

The log is for reading over your own shoulder while something goes wrong: prose,
second resolution, and a rolling buffer that truncates to zero at 1 MB. The
questions worth asking here are asked over hundreds of dictations at once — which
words is the model least sure of, does a missing ending come from the decoder or
the gate, what does a prompt stage really cost on your own sentences — and none
of them survive a file that throws the oldest away. **No size cap on this one**,
deliberately: a year of heavy use is a few megabytes.

`docs/cli.md` carries a `jq` recipe for each of those questions.

## Decisions worth arguing with

- **No config key.** Recordings are kept unconditionally and a clip is strictly
  more revealing than its transcript, so a switch on the trace alone protects
  nothing.
- **A task local, not a parameter.** Threading a collector through
  `Pipeline.run`, `apply` and `runPrompt` for a debug artefact was the invasive
  version; a dictation is one task from the decoder to the last stage, which is
  exactly the scope a task local has.
- **A word keeps its lowest token's confidence, not the mean.** A name that came
  through as one solid piece and one shaky one is a shaky name. FluidAudio's own
  `buildWordTimings` drops confidence entirely, so the grouping is ours.
- **`--transcribe` writes a line too**, tagged `cli` against `live`. This is what
  makes the file worth keeping: change a table, re-run the archive, and both sets
  of numbers sit together joined to the same audio.

## Verification

Run over real clips — word timings and confidences populate, and every `jq`
recipe in `cli.md` was run against the file before being written down.

```
check-pipeline       27/27
check-replacements   23/23
check-numbers        97/97
check-dotted         54/54
```

Unchanged, and the check scripts bind no collector, so `--pipeline` runs leave
the corpus clean. The 188 pre-existing Swift 6 sendability warnings are untouched.

**Not yet verified:** the live hotkey path. It builds and runs, but confirming it
needs someone to press the key and check
`~/Recordings/ParrotFlow\ Dev/trace.jsonl`.

## Left alone

`Log.swift`'s 1 MB cap, which truncates the whole log to zero rather than
rotating. Out of scope here, but worth a follow-up.

---

The first line this wrote was already a finding: a clip with 1.22s of speech
across one VAD segment that decoded to nothing at confidence 0.1. The log showed
the speech, and then silence.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -651,7 +651,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 // decoder: the stages that follow write into the same record,
                 // and the line is appended even if one of them throws.
                 let text = try await Trace.record(
-                    wav: recording.url.lastPathComponent, source: .live, app: app?.name
+                    wav: recording.url.lastPathComponent, source: .live,
+                    app: app.map { Trace.App(name: $0.name, bundleID: $0.bundleID) }
                 ) {
                     let text = try await self?.transcriber.transcribe(
                         url: recording.url, config: config, app: app,
@@ -1254,6 +1255,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             do {
                 try ConfigWriter.addReplacement(heard: rule.heard, corrected: rule.corrected)
                 Log.write("learned replacement: \(rule.heard) -> \(rule.corrected)")
+                Trace.correction(heard: rule.heard, corrected: rule.corrected, via: "command")
             } catch {
                 presentAlert(title: "Could not save the rule", message: error.localizedDescription)
                 pendingSelection = nil
@@ -1642,6 +1644,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 do {
                     try ConfigWriter.addReplacement(heard: rule.heard, corrected: rule.corrected)
                     Log.write("learned replacement: \(rule.heard) -> \(rule.corrected)")
+                    Trace.correction(
+                        heard: rule.heard, corrected: rule.corrected, via: "panel"
+                    )
                 } catch {
                     self.presentAlert(
                         title: "Could not save the rule", message: error.localizedDescription
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -652,7 +652,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 // and the line is appended even if one of them throws.
                 let text = try await Trace.record(
                     wav: recording.url.lastPathComponent, source: .live,
-                    app: app.map { Trace.App(name: $0.name, bundleID: $0.bundleID) }
+                    app: app.map { Trace.App(name: $0.name, bundleID: $0.bundleID) },
+                    beside: recording.url.deletingLastPathComponent()
                 ) {
                     let text = try await self?.transcriber.transcribe(
                         url: recording.url, config: config, app: app,
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -213,6 +213,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // which the app never runs. A mistyped stage name used to vanish at
         // decode time with no trace anywhere: replacements simply stopped
         // happening, and nothing distinguished that from an empty table.
+        // Beside the recordings, and re-read here because `output_dir` can move
+        // on any save of config.yaml.
+        Trace.directory = config.resolvedOutputDir
+
         configProblems = config.problems()
         for problem in configProblems { Log.write("config: \(problem)") }
         // Logged and not flashed. A `command:` transform is announced on every
@@ -643,18 +647,27 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 // "Transcribing…" is the truth until the decoder is done, and
                 // a lie for the second a prompt or a script spends after it.
                 // A stage that wrote a `display:` says so itself.
-                let text = try await self?.transcriber.transcribe(
-                    url: recording.url, config: config, app: app,
-                    progress: { label in
-                        Task { @MainActor [weak self] in
-                            // Still this dictation, and still one that has
-                            // something on screen to replace.
-                            guard let self, self.transcriptionRun == run,
-                                  self.transcriptionLabel != nil else { return }
-                            self.beginProgress(label)
+                // The trace is opened around the whole chain, not just the
+                // decoder: the stages that follow write into the same record,
+                // and the line is appended even if one of them throws.
+                let text = try await Trace.record(
+                    wav: recording.url.lastPathComponent, source: .live, app: app?.name
+                ) {
+                    let text = try await self?.transcriber.transcribe(
+                        url: recording.url, config: config, app: app,
+                        progress: { label in
+                            Task { @MainActor [weak self] in
+                                // Still this dictation, and still one that has
+                                // something on screen to replace.
+                                guard let self, self.transcriptionRun == run,
+                                      self.transcriptionLabel != nil else { return }
+                                self.beginProgress(label)
+                            }
                         }
-                    }
-                ) ?? ""
+                    ) ?? ""
+                    Trace.current?.recordFinal(text)
+                    return text
+                }
                 await MainActor.run {
                     guard let self else { return }
                     // Only if the screen is still ours. Push-to-talk does not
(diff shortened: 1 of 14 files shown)


znat:
All three addressed in e48e551 — checked on merit first, and all three were real.

**1. Cross-process appends.** Valid. An append was a `seekToEnd` then a `write`, and `FileHandle(forWritingTo:)` does not open with `O_APPEND`. Two processes write this file by design — the app, and the `--transcribe` archive sweep `cli.md` recommends, which nobody will quit the app before starting — so both can seek before either writes. Now opened `O_WRONLY | O_APPEND | O_CREAT` with a single `write(2)`, which is the one construction the kernel will not interleave. The serial queue stays; it orders this process's writes and keeps them off the caller's thread, and never had anything to say about the other process.

Worth being straight about the evidence: **I could not reproduce corruption before the fix.** 20 simultaneous `--transcribe` processes produced 20 valid lines on the pre-fix build. Each process writes exactly once and model-load jitter spreads the writes out, so the window is narrow. Narrow is not closed, and the fix costs nothing — but I'm not going to claim I caught it in the act.

**2. Config reload misplaces records.** Valid, and narrow, and the fix is strictly better design anyway. The directory is now captured when the dictation starts rather than read from the global when it ends. `output_dir` can change on any save of `config.yaml` — the file is watched and reloaded live — and a save while a prompt stage was running would file the record in the new directory while the clip it names sits in the old. `wav` is a bare filename, so separating those breaks the join the entire format rests on.

**3. Failure records abort jq.** Valid, and the most embarrassing of the three: the writer deliberately emits records with `asr: null` when a dictation throws before the decoder returns, and then the documented recipe iterates `.asr.words[]` over them. jq doesn't skip — it aborts with exit 5. So the one query you'd run after a bad day was the one that couldn't run. Confirmed before fixing:

```
$ jq -r '.asr.words[] | select(.confidence < 0.5) | .word' n.jsonl
jq: error (at n.jsonl:1): Cannot iterate over null (null)     # exit 5
$ jq -r '.asr.words[]? | select(.confidence < 0.5) | .word' n.jsonl
                                                              # exit 0
```

Took the `[]?` suggestion, with a comment saying why. The other three recipes were already null-safe — recipe 2's `select(.asr.words|length > 0)` filters those records out, since `null | length` is 0.

**Verification:** 20 concurrent writers → 20 valid lines. The fixed recipe returns its answer over a file containing a null-`asr` record. check-pipeline 27/27, check-replacements 23/23, check-numbers 97/97, check-dotted 54/54.

Still unverified, as in the description: the live hotkey path.

znat:
Pushed `8c48462` — schema v2, folded in here rather than as a follow-up so the first records ever written are v2 instead of v1-then-migrate.

**Added**, by one rule — log what is unrecoverable, compute the rest:

| Field | Why it can't wait |
|---|---|
| `v: 2` | Free before there is anything to migrate. Existing lines have no `v` and read as 1 |
| `kind` + correction lines | The only free human labels in the system; unreconstructable afterwards |
| `lang` | Ours, not Parakeet's — see below |
| `app.bundle_id` | `app:` conditions match name+id joined; only the name was kept |
| `asr.model` | Otherwise a moved metric can't be attributed to prompt vs model |
| `skip_reason` | Category beside the prose; the prose can't be grouped |

**Not added, deliberately:** pause gaps, filled pauses, confidence dips, gate-would-have-fired. All computable from `asr.words` + `vad.segments`; freezing them in means recomputing history every time a threshold moves.

**Two findings that shaped this**, both from real audio in `~/Recordings`:

*Parakeet does not strip disfluency.* `Um` (conf 0.910) and `uh` (conf 0.529) both come out of the decoder. What removes them is `config.yaml:104`. That's only visible because `asr.text` is captured before the tables run — and the stage-rewrite recipe now shows it: `"Um currently in the logs…"` → `"Currently in the logs…"`.

*Self-correction is acoustically invisible.* On a clip where the speaker said "but do we feel do you do we keep also", every word is 0.98+ confidence and the gaps around the false start (0.16s, 0.24s) are **smaller** than gaps in fluent stretches of the same sentence (0.64s, 0.96s). Genuine pausing does show up — a 3.92s gap corroborated by VAD going quiet. So hesitation is acoustic and already derivable; going back is semantic and no timing field will ever catch it.

**On `lang`:** `ASRResult` has no language field, and `TokenLanguageFilter` takes a language *hint* and filters by script — it is not a detector. So there is nothing to log from the model. `Pipeline.forText` has always computed the verdict that selects the pipeline and `Replacements.apply` discarded it on the same line; that's what `lang` now carries, and it's the more useful one since it's what decided what ran.

**On `edit`:** rejected as designed. Catching post-paste edits means reading another app's content back through Accessibility after focus has been given away. The corrections ParrotFlow already mediates — panel, spoken command, `--learn` — are free and carry the same signal.

**Verification.** All six documented jq recipes exit 0 against a file holding dictations, a correction line and a legacy null-`asr` record together. `check-pipeline` 27/27, `check-replacements` 23/23, `check-numbers` 97/97, `check-dotted` 54/54, `check-wake` 24/24. `check-routing` is 50/54 — I checked the baseline on `main` and it was already 50/54, so not a regression.

Still unverified: the live hotkey path.

znat:
Second review addressed in `82897bb` — apologies, I pushed v2 without re-reading the PR and missed this round.

**1. Trace destination captured too late — valid, and the fix was the wrong shape.**

Right that moving the read earlier only narrowed the window. `output_dir` can change on any save of `config.yaml`, so *any* read of the global happens at some particular instant: reading it when the record finishes loses to a save during a prompt stage; reading it when the dictation starts loses to a save between the clip being written and the task being scheduled. There is no instant that is both late enough to be right and early enough to be safe — which is the tell that a global was the wrong source.

It no longer reads one. The line is filed beside `recording.url.deletingLastPathComponent()` — where the file actually went, whatever the config said then or says now. No snapshot, no window.

That surfaced a related hole you didn't raise: `--transcribe` on a clip outside the recordings directory logged a bare filename that wasn't in the directory the line landed in, so the join was already broken there. The line still goes to the recordings directory (one corpus in one place; `--transcribe /tmp/t.wav` shouldn't leave a `trace.jsonl` in /tmp), so a clip from outside is now named by its full path. Verified both ways:

```
in corpus:     parrotflow-2026-07-30T13-02-15.wav
outside:       /Users/…/Recordings/ParrotFlow Dev/parrotflow-2026-08-04T09-39-12.wav
```

**2. Missing VAD aborts analysis — right conclusion, wrong mechanism.**

`.vad.segments[-1][1]` does **not** abort jq when `vad` is null. jq returns null for *indexing* null; it only errors on *iterating* it, which is why the earlier `[]?` fix was necessary and this is a different case. Tested directly:

```
$ jq -r '.vad.segments[-1][1]' vnull.jsonl      # null,  exit 0
$ jq -r '.vad.segments[]'      vnull.jsonl      # Cannot iterate over null, exit 5
$ <old recipe 2>                                # gate-off.wav\t\t0.2\t   exit 0
```

So the shipped recipe emits a row of blanks rather than aborting — worse in a different way, since it quietly pollutes an average instead of stopping to complain. Took the guard for that reason.

The premise underneath is sound, though, and worth stating: `vad` can be null with every word present via three routes — `audio.speech_gate` off, the detector failing to load, or the channel failing to read.

**Verification.** All six documented recipes exit 0 over a file holding a dictation, a correction, a legacy null-`asr` record and a null-`vad` record together. check-pipeline 27/27, check-replacements 23/23, check-numbers 97/97, check-dotted 54/54.

PR #41, opened 2026-08-04
feat: a transform is a folder, and it brings its own things with it

znat:
A transform is a directory. Everything belonging to it — the prompt or the script, its case set, the roster it looks names up in — lives in `transforms/<name>/` beside the config that declared it.

```
~/.config/parrotflow/
  config.yaml
  transforms/
    code_identifiers/
      code_identifiers.py     # the entry point, named after the transform
      cases.yaml              # its set, right there
    slack_mentions/
      slack_mentions.py
      roster.json             # data it owns
```

That folder is also **the working directory the command runs in**, which is the half that pays for the extra directory: a script opens `roster.json` by that name, so the folder is a thing you can copy to another machine and have work.

Phase 1 of `docs/proposals/transform-folders.md`, which is in this diff. Phase 2 — the `--eval` harness — follows in a stacked PR and needs this one.

## Nobody's setup changes because they upgraded

The old location beside `config.yaml` is searched second. A script there resolves, runs, and gets one line from `--check-config`:

```
· transforms: "code_identifiers" found at the old location —
    move it to transforms/code_identifiers/code_identifiers.py
```

A notice, through `notices()` and not `problems()` — it exits 0. Nothing is moved for you, and seeding refuses to write a fresh script into the folder when an edited one already exists beside the config, which would otherwise put a new file *in front* of your tuned stop lists silently.

## Three decisions worth checking

**`prompt:` and `replace:` gained `{ path: … }`, and a scalar stays a scalar.** An inline `prompt: |` is unchanged — a three-line prompt is worse in a file than in the config it belongs to. The mapping form is tried first and only matches `path: <string>`, so a table whose one key happens to be `path` is still a table, and a malformed table still throws where it always did.

**A `path:` naming nothing takes out that transform and nothing else.** It is skipped, `--check-config` says why with a ✗, and the rest of the config goes on working.

**`--check-config` now prints the resolved absolute path of every body.** A file present both in the folder and beside `config.yaml` was otherwise invisible, and "which one is running" is the first question when an edit had no effect.

## Two additions the proposal did not ask for

Both exist because they turn assertions into checks: `PARROTFLOW_CONFIG_DIR` points any command at a config directory of its own, and `--seed-config` reports what a first launch wrote and what it left alone. Together they let `scripts/check-transform-folders.sh` score the real binary against a config in `/tmp` instead of inheriting whatever the machine happens to have.

## Also in here

The repository moved to match: `examples/transforms/<name>/` holds the shipped transforms and the sets that score them, a byte-for-byte mirror of what a first launch writes, kept honest by `check-seeded-transform.sh` — which has never run in CI under either name until now.

And the config comments are cut back to what a field does, with the reasoning left in `docs/`. `Config.defaultYAML` had grown to 264 comment lines, more prose than `config.example.yaml`, which is meant to be the fuller of the two. Both files parse to structures identical to before; that is checked, not asserted.

One bug surfaced on the way: `expandingTildeInPath` also standardises, so `command: sed -e s/quick/slow/` lost its trailing slash before the shell ever saw it.

## Upgrading is one-way

A config that uses `prompt: { path: … }` cannot be read by 0.4.x — the app refuses the whole file, not just that entry. Worth a line in the release notes.

## Verification

| harness | result |
|---|---|
| `check-transform-folders.sh` (new, 21 cases) | 21/21 |
| `check-seeded-transform.sh` | both files identical to `examples/` |
| `check-default-config.sh` | 27 keys, 9 transforms, 2 steps |
| `check-dotted.sh` | 54/54 + 2 known-unfixable |
| `check-replacements.sh` | 23/23 |
| `check-numbers.sh` | 97/97 |
| `swiftlint` | 0 errors |

Fresh install, upgrade-with-old-script, and both spellings of a relative path are all covered by the new set rather than by hand.


znat:
## The review finding is valid, and fixed in fbd16d4

Greptile is right, and it broke the one promise this PR makes. Worth spelling out because the shape is not obvious:

The working directory was chosen from the folder, and the program from the search path, **independently**. On an upgraded install those two disagree — because seeding writes `transforms/<name>/cases.yaml` even when the script stays beside `config.yaml`. So the folder exists, the program is not in it, and a script that has read `roster.json` from beside `config.yaml` for months gets started in a directory that never held one. The open fails, the command exits non-zero, and a non-zero command keeps the transcript: the stage quietly stops doing anything.

The fix is that the program decides. A script in the folder opens its neighbours in the folder; a script at the old location keeps the neighbours it has always had. `TransformFolder.workingDirectory` is now only the answer for a command with no file of its own — a bare `sed` off PATH.

`check-transform-folders.sh` gained the case, built in exactly that shape: folder present, script beside the config, data file only the old location has. It reports **21/22 on the code this replaces** and 22/22 with it, so it is a case that would have caught this rather than one written to match the fix.

## On the red `checks` run

`check-pipeline.sh` failed one case — `a command transform keeps what it was given intact` — and **the same commit passed on re-run with no change**, so it is flaky rather than broken. Two things point away from this PR:

- The verbose re-run *inside the same failing job* produced the correct output for the same input, moments later.
- 40/40 locally, and the diff to `CommandRunner` is the signature, the working directory and path resolution — the read/wait handoff is untouched.

The symptom is the whole of stdout going missing, which returns nil, which keeps the transcript. That points at the existing `readabilityHandler` / `readDataToEndOfFile` handoff losing a chunk under load rather than at anything here — but one green re-run is not a proof, and I would rather say so than call it fixed. Worth its own change if it recurs.

znat:
## Also valid, and fixed in 28b2e92

This is the same bug one layer down, and my first fix did not reach it. Worth being precise about why.

`fbd16d4` took the working directory from the resolved **program**. That is the right answer for `command: legacy.py` and no answer at all for `command: python3 legacy.py` — there the program is an interpreter the shell finds on PATH, so nothing about the program says where the transform lives. Only the argument does. An upgraded config written that way was still being started in a folder its script has never been in, and still failing the same silent way: the interpreter cannot find the file, the command exits non-zero, and a non-zero command keeps the transcript.

The rule is now **the directory of the first file the command names**, program or argument. One sentence covering both shapes, with the folder as the fallback only for a command that names no file at all — `sed`, `tr` — which reads nothing and does not care.

Only the working directory is decided that way. `parts` still reports the *program*, deliberately: that is what `complaint` checks the execute bit of, and `python3 legacy.py` needs no execute bit on `legacy.py`.

## Two cases, not one

The obvious fix for the reported shape would break the shape this PR exists to enable, so both are covered — the same script wrapped in `python3`, once at the old location and once in its folder:

```
✗ an interpreter-wrapped script at the old location still runs      # 23/24 before
✓ an interpreter-wrapped script in the folder reads its own neighbours
```

`check-transform-folders.sh` is 24/24 with the fix and 23/24 without it, so the new case fails on the code it replaces rather than being written to match the fix.

`docs/pipelines.md` and `docs/authoring.md` said "the folder is the working directory", which is no longer the rule; both now state it as the first file the command names.

znat:
## Valid, and fixed in 103e9ee — but it found a worse mistake underneath

The reported case is real: splitting the arguments on every space misses `python3 'my scripts/x.py'`, where the quotes are there precisely because the path has a space in it. That is the same problem `parts` already solves for the program, and it has the same answer — something has to say which spaces separate, and here it is the quoting, because YAML hands the string over verbatim. So the arguments are now split the way the shell will split them.

Fixing that surfaced the actual bug, which was in my previous fix rather than in the split:

```
can't open file '…/my scripts/my scripts/spaced.py'
```

**Aiming the working directory at the file's own directory is right for the program and wrong for an argument.** The program is rewritten to an absolute path before the shell sees it, so moving the working directory cannot disturb it. An argument arrives exactly as written and is resolved against whatever we set — so pointing at the script's directory counts the relative path twice.

So an argument now gives the directory it was found **under**, which is what the path is relative to, and also what the command ran in before folders existed. `TransformFolder.Resolved` carries that base explicitly, because "where it lives" and "what it is relative to" are different questions and the code was answering one of them twice.

## One correction to my own case

The case I added in 28b2e92 asserted that a spaced-path script reads `suffix.txt` from beside *itself*. That was never the behaviour — under the old code the working directory was the config directory, so that script read `suffix.txt` from there. The case now puts the file where it has always been and asserts the command still runs where it always ran, which is the thing this PR must not break. Thank you for the push that exposed it; the wrong assertion would have shipped as a "verified" one.

`check-transform-folders.sh` is 25/25, and the quoted case fails 24/25 on the code it replaces.

## On this family of findings

Three rounds, all the same shape: *what does the working directory have to be so that an existing config keeps working?* The rule now reads as one sentence — the directory of the first file the command names, taken as the file's own directory for the program and as the search base for an argument — and the three cases sit in the set together, so the next variant has something to fail against.

znat:
## The fallback is gone, and with it this whole family of findings

Four review rounds, all the same question: *which of two directories does this command run in?* The answer is that there should never have been two.

The old-location fallback was protecting installs that do not exist. Three releases, **six downloads between them**, all of them my own `try-install` runs — one star, no forks. I was maintaining, reviewing and arguing over a migration path for a population of zero, and every finding above was a different config where "which directory" had a different right answer.

It also makes the proposal's own headline decision true. Its table says **"One layout, enforced. Folders. There is no flat alternative to fall back to"** — and then the Back-compat section three below specified one. Those were always in tension.

## What changed

There is one place a transform's files are: its folder. Either spelling still names the same file — `slack_mentions.py` and `transforms/slack_mentions/slack_mentions.py` — because that is one place reached two ways, and the second is accepted **only when it lands inside the folder**, so nothing outside is ever reachable.

Replacing the fallback: one line from `--check-config`, and a fault rather than a notice.

```
✗ transforms: "code_identifiers" — code_identifiers.py is not in
    transforms/code_identifiers/, and the shell cannot find it either
```

Left to the shell that failed once per transcript into the log while the pipeline returned the text unchanged. Said here, it is answerable by moving one file.

## What came out

| | |
|---|---|
| `workingDirectory(for:in:)` + `words(of:)` | ~70 lines, and all four findings |
| `atOldLocation`, `Resolved.base`, `moveNotice` | nothing left to compare |
| the old-location notice in `notices()` | — |
| seeding's "preserve the legacy script" branch | — |
| 6 back-compat check cases | replaced by 2 about a path that is not there |

**16 files, +194 −466.** `check-transform-folders.sh` is 12/12 and covers more than the 25 cases did.

## Two fixtures had stopped testing what they claim

Moving the `--pipeline` fixtures into folders showed that `command-hangs` and `command-unrunnable` expect the transcript back unchanged — which is also exactly what a script nobody can find produces. They were passing on the fallback, and would have passed on nothing at all. Both now resolve properly and fail for their own reasons.

`command-relative` reaches the shipped example through a symlink, so it still scores the file people actually edit.

## Verification

All eleven deterministic sets green, swiftlint clean, and both real configs on this machine converted and reporting `exit 0`.

PR #43, opened 2026-08-04
feat: the harness that scores a rewrite ships with the app

znat:
Stacked on #41 — review that first; this diff is only `--eval`. GitHub retargets this to `main` when #41 merges.

Every rewrite in this repository has a case set and a bespoke runner in `scripts/`. Nobody who installed the app has either, which leaves "write a prompt and hope" as the only loop available to them. This is the loop `docs/authoring.md` prescribes, in the binary, against your own config.

```sh
$PF --eval code_identifiers                        # transforms/…/cases.yaml
$PF --eval slack_mentions --cases heldout.yaml     # another set in the folder
$PF --eval grammar --probe restraint --verbose
$PF --eval ~/scratch/my-cases.yaml                 # one-off, outside a folder
```

It resolves the transform from your config and runs **that** — a one-step `Pipeline` and `PromptRunner`, the same code the app runs, including the part where every way of failing leaves the text alone. A runner that reimplements the thing it scores drifts from it, and the number then describes code nobody ships.

## The format

`input` and `expect` are the whole requirement.

```yaml
# The contract, in prose: what counts as a case here and what is deliberately
# out of scope. It is what you will disagree with yourself about in a week.
cases:
  - probe: ambiguity_common
    input:  mark it as resolved
    expect: mark it as resolved
```

**`expect` left out means "comes back exactly as it went in"** — detected, never declared, and reported as its own half. A rewrite that scores well on `change` and badly on `keep` is not 80% of the way there; it is one that makes you proof-read every dictation.

Four optional keys, each of which caught something real while building `scripts/validate-slack-mentions.py`:

| | |
|---|---|
| `probe:` | the breakdown. One broken category hides behind eleven healthy ones for exactly as long as nobody breaks it out |
| `instruction:` | what the speaker said, for a prompt reached by voice — see below |
| `control:` | the same set without a model. The only thing that answers "is the model earning its place" |
| `intermediate:` | `field:`, `resolve:`, optionally `produce:`. Scores what the first stage alone returns, so you can tell whether the prompt or the code is at fault |

Where an intermediate gold exists it is **checked against itself before anything is scored**: resolving it must produce `expect`. A typo in a gold otherwise scores every candidate against the typo, silently, for as long as the set exists — so that is a refusal, not a warning.

Every case lands in a probe bucket including the ones that name none, because a grid that silently covers 61 of 75 reads as a complete breakdown and is not one. Latency is warm, after a throwaway first call.

## Both shipped examples agree with the runners they replace

| | `--eval` | the bespoke runner |
|---|---|---|
| `code_identifiers`, 75 cases | 38/43 change, 30/32 keep, **68/75** | `validate-code-identifiers.py --code-only`: identical |
| `slack_mentions`, 76 cases | 40/40, 36/36, **76/76** | `validate-slack-mentions.py --code-only`: identical |
| `slack_mentions`, held out | 10/10, 13/14, **23/24** | identical, same single failure |
| `grammar`, 17 cases | 7/7, 9/10, **16/17** | `check-grammar.sh`: identical |

`grammar` only matches **with** `instruction:` in its set. Run as a pipeline stage the same prompt is given no instruction and comes back at 15/17 — a real number for a use that transform does not have. That is why the key exists, and it is now recorded in the case file rather than in a runner's argv.

## Two decisions worth checking

**`--eval` exits 0 whenever it managed to score, whatever the number.** A set worth having keeps its residue in, failing — `dotted` scores 54/54 and carries two more it cannot do. A number below 100% is an ordinary result, not something to report to a shell. It exits 1 when it could not score at all: no such transform, no case file, a bad gold.

**Custom gates stay bespoke.** The Slack set counts wrong handles and false `@channel` separately and both must be zero; that does not generalise, and a generic runner that tried would be guessing at the one number that exists to be trusted.

## Verification

`scripts/check-eval.sh` is new — 13 cases, no model anywhere in it, so the numbers are the same on a runner and on a laptop with Ollama up. It covers all six scoring requirements, both halves, the per-probe grid, the control, the intermediate, and the bad-gold refusal, including that **nothing is scored before the refusal**.

| harness | result |
|---|---|
| `check-eval.sh` (new) | 13/13 |
| `check-transform-folders.sh` | 12/12 |
| the other seven deterministic sets | unchanged |
| `swiftlint` | 0 errors |

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -117,6 +117,12 @@ jobs:
       - name: A folder per transform
         run: scripts/check-transform-folders.sh
 
+      # The scoring harness itself, against sets built for the purpose. No
+      # model anywhere in it: the numbers have to be the same on a runner and
+      # on a laptop with Ollama up.
+      - name: The eval harness
+        run: scripts/check-eval.sh
+
       # Not a validation set: it checks that the file a new install is written
       # with still teaches everything config.example.yaml documents. That has
       # drifted twice, both times silently.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,15 +22,19 @@ $PF --check-config                                  # what will actually run
 $PF --pipeline <file.yaml> "<text>" --app <name>    # a stage, without a mic
 $PF --replace "<text>"                              # the substitution tables
 $PF --route "<what someone would say>"              # which transform that reaches
+$PF --eval <transform>                              # score it against its set
 ```
 
 `--check-config` reports what survived parsing, which is not what the file
 says. Everything else is in [docs/cli.md](docs/cli.md).
 
-**Never change a prompt or a pattern without scoring it.** Every rewrite in
-this repo has a case set in `tests/` and a runner in `scripts/`. Get the number
-before, change one thing, get it after. "It looks better" is how a change fixes
-the example in front of you and breaks two you are not looking at.
+**Never change a prompt or a pattern without scoring it.** Every rewrite has a
+case set — a transform's in its own folder under `examples/transforms/<name>/`,
+everything else in `tests/` — and a runner, either `$PF --eval <name>` or a
+script in `scripts/`. Get the number before, change one thing, get it after.
+"It looks better" is how a change fixes the example in front of you and breaks
+two you are not looking at. Read both halves: `keep` is the one that decides
+whether a rewrite is usable.
 
 **Fail open.** A stage that errors, times out or finds no model must leave the
 transcript exactly as it arrived. Test it by stopping Ollama and running the
--- a/Sources/ParrotFlow/Config.swift
+++ b/Sources/ParrotFlow/Config.swift
@@ -85,9 +85,12 @@ struct Config: Decodable, Equatable {
         var namesBoth = false
         /// A `path:` that named a file this could not read, in words.
         var unreadable: String?
+        /// `tests: { path: heldout.yaml }` — the set `--eval` scores by default.
+        var tests: String?
 
         enum CodingKeys: String, CodingKey {
             case name, description, display, confirm, prompt, content, replace, command
+            case tests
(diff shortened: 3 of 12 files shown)


znat:
## Rebased onto `main` now that #41 has merged, and all three findings fixed in 075292f

The first one was the serious one.

**The set's own transform was being ignored.** `locate` prefers a `transforms:` section carried by the case file — that is how a set states what it assumes and scores the same anywhere. `through` then appended it to the config *only when no entry of that name was already there*, handing the decision back to the machine. Silently, and only on the machines where the two differ, which is precisely the case the set carried its own definition to survive. The resolved transform now replaces whatever the config had, because `locate` already decided.

**A `--probe` that matched nothing scored nothing and exited 0.** The documented contract is that 0 means "this was scored", so a typo in a probe name read as a clean run. Refused now, and it says which probes the set actually has.

**A `tests:` that is neither form was ignored.** `tests: { file: heldout.yaml }` decoded as nothing at all, and the transform then scored `cases.yaml` while the config said otherwise. A key that does not do what it says is worse than one that fails, so it is skipped with a reason like any other unreadable body.

`scripts/check-eval.sh` gained four cases, and all four fail on the code they replace — **13/17 before, 17/17 after**.

## Still true after the rebase

The acceptance numbers are the point of this PR, so they were re-run against the real transforms on this machine rather than assumed:

| | `--eval` | the bespoke runner |
|---|---|---|
| `code_identifiers` | 68/75 | identical |
| `slack_mentions` | 76/76 | identical |
| `slack_mentions --cases heldout.yaml` | 23/24 | identical, same single failure |

All eleven deterministic sets green, swiftlint clean.

znat:
Closed by GitHub when its base branch `feat/transform-folders` was deleted on #41's merge. Reopened as #44, rebased onto `main`, with the three findings above fixed.

PR #44, opened 2026-08-04
feat: the harness that scores a rewrite ships with the app

znat:
Supersedes #43, which GitHub closed when its base branch was deleted on #41's merge. Rebased onto `main`; the diff is only `--eval`.

Every rewrite in this repository has a case set and a bespoke runner in `scripts/`. Nobody who installed the app has either, which leaves "write a prompt and hope" as the only loop available to them. This is the loop `docs/authoring.md` prescribes, in the binary, against your own config.

```sh
$PF --eval code_identifiers                        # transforms/…/cases.yaml
$PF --eval slack_mentions --cases heldout.yaml     # another set in the folder
$PF --eval grammar --probe restraint --verbose
$PF --eval ~/scratch/my-cases.yaml                 # one-off, outside a folder
```

It resolves the transform from your config and runs **that** — a one-step `Pipeline` and `PromptRunner`, the same code the app runs, including the part where every way of failing leaves the text alone. A runner that reimplements the thing it scores drifts from it, and the number then describes code nobody ships.

## The format

`input` and `expect` are the whole requirement.

```yaml
# The contract, in prose: what counts as a case here and what is deliberately
# out of scope. It is what you will disagree with yourself about in a week.
cases:
  - probe: ambiguity_common
    input:  mark it as resolved
    expect: mark it as resolved
```

**`expect` left out means "comes back exactly as it went in"** — detected, never declared, and reported as its own half. A rewrite that scores well on `change` and badly on `keep` is not 80% of the way there; it is one that makes you proof-read every dictation.

Four optional keys, each of which caught something real while building `scripts/validate-slack-mentions.py`:

| | |
|---|---|
| `probe:` | the breakdown. One broken category hides behind eleven healthy ones for exactly as long as nobody breaks it out |
| `instruction:` | what the speaker said, for a prompt reached by voice — see below |
| `control:` | the same set without a model. The only thing that answers "is the model earning its place" |
| `intermediate:` | `field:`, `resolve:`, optionally `produce:`. Scores what the first stage alone returns, so you can tell whether the prompt or the code is at fault |

Where an intermediate gold exists it is **checked against itself before anything is scored**: resolving it must produce `expect`. A typo in a gold otherwise scores every candidate against the typo, silently, for as long as the set exists — so that is a refusal, not a warning.

Every case lands in a probe bucket including the ones that name none, because a grid that silently covers 61 of 75 reads as complete and is not. Latency is warm, after a throwaway first call.

## Both shipped examples agree with the runners they replace

| | `--eval` | the bespoke runner |
|---|---|---|
| `code_identifiers`, 75 cases | 38/43 change, 30/32 keep, **68/75** | `validate-code-identifiers.py --code-only`: identical |
| `slack_mentions`, 76 cases | 40/40, 36/36, **76/76** | `validate-slack-mentions.py --code-only`: identical |
| `slack_mentions`, held out | 10/10, 13/14, **23/24** | identical, same single failure |
| `grammar`, 17 cases | 7/7, 9/10, **16/17** | `check-grammar.sh`: identical |

`grammar` only matches **with** `instruction:` in its set. Run as a pipeline stage the same prompt is given no instruction and comes back at 15/17 — a real number for a use that transform does not have. That is why the key exists, and it now lives in the case file rather than in a runner's argv.

## Two decisions worth checking

**`--eval` exits 0 whenever it managed to score, whatever the number.** A set worth having keeps its residue in, failing — `dotted` scores 54/54 and carries two more it cannot do. It exits 1 when it could not score at all: no such transform, no case file, a bad gold, or a `--probe` that matches nothing.

**Custom gates stay bespoke.** The Slack set counts wrong handles and false `@channel` separately and both must be zero; that does not generalise, and a generic runner that tried would be guessing at the one number that exists to be trusted.

## Three findings from the review of #43, fixed here

- **The set's own transform was being ignored** when the machine's config had an entry of the same name — silently, and only where the two differ, which is the case a set carries its own definition to survive.
- **A `--probe` that matched nothing** scored nothing and exited 0, so a typo read as a clean run.
- **A malformed `tests:`** (`{ file: … }` rather than `{ path: … }`) decoded as nothing, and the transform then scored `cases.yaml` while the config said otherwise.

All three have cases, and all four assertions fail on the code they replace.

## Verification

`scripts/check-eval.sh` is new — 17 cases, no model anywhere in it, so the numbers are the same on a runner and on a laptop with Ollama up.

| harness | result |
|---|---|
| `check-eval.sh` (new) | 17/17 |
| `check-transform-folders.sh` | 12/12 |
| the other seven deterministic sets | unchanged |
| `swiftlint` | 0 errors |

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -117,6 +117,12 @@ jobs:
       - name: A folder per transform
         run: scripts/check-transform-folders.sh
 
+      # The scoring harness itself, against sets built for the purpose. No
+      # model anywhere in it: the numbers have to be the same on a runner and
+      # on a laptop with Ollama up.
+      - name: The eval harness
+        run: scripts/check-eval.sh
+
       # Not a validation set: it checks that the file a new install is written
       # with still teaches everything config.example.yaml documents. That has
       # drifted twice, both times silently.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,15 +22,19 @@ $PF --check-config                                  # what will actually run
 $PF --pipeline <file.yaml> "<text>" --app <name>    # a stage, without a mic
 $PF --replace "<text>"                              # the substitution tables
 $PF --route "<what someone would say>"              # which transform that reaches
+$PF --eval <transform>                              # score it against its set
 ```
 
 `--check-config` reports what survived parsing, which is not what the file
 says. Everything else is in [docs/cli.md](docs/cli.md).
 
-**Never change a prompt or a pattern without scoring it.** Every rewrite in
-this repo has a case set in `tests/` and a runner in `scripts/`. Get the number
-before, change one thing, get it after. "It looks better" is how a change fixes
-the example in front of you and breaks two you are not looking at.
+**Never change a prompt or a pattern without scoring it.** Every rewrite has a
+case set — a transform's in its own folder under `examples/transforms/<name>/`,
+everything else in `tests/` — and a runner, either `$PF --eval <name>` or a
+script in `scripts/`. Get the number before, change one thing, get it after.
+"It looks better" is how a change fixes the example in front of you and breaks
+two you are not looking at. Read both halves: `keep` is the one that decides
+whether a rewrite is usable.
 
 **Fail open.** A stage that errors, times out or finds no model must leave the
 transcript exactly as it arrived. Test it by stopping Ollama and running the
--- a/Sources/ParrotFlow/Config.swift
+++ b/Sources/ParrotFlow/Config.swift
@@ -130,8 +130,18 @@ struct Config: Decodable, Equatable {
             // `tests: { path: heldout.yaml }` mean the same thing, because
             // having to remember which of the two a key takes is the kind of
             // thing a config format should not ask.
-            let testsWritten = (try? trimmed(.tests)) ?? ""
-            tests = path(.tests) ?? (testsWritten.isEmpty ? nil : testsWritten)
+            // Written either way, and refused when it is neither. A mapping
+            // that is not `{ path: … }` — `tests: { file: heldout.yaml }` —
+            // used to decode as nothing at all, and the transform then scored
+            // `cases.yaml` while the config said otherwise. A key that does
+            // not do what it says is worse than one that fails.
+            if let written = path(.tests) {
+                tests = written
+            } else if let scalar = try? trimmed(.tests) {
+                tests = scalar.isEmpty ? nil : scalar
+            } else {
+                unreadable = "`tests:` is neither a filename nor `{ path: <filename> }`"
+            }
 
             // A body written either way. The mapping is tried first and only
             // succeeds on `{ path: <string> }`, so nothing that decoded before
--- a/Sources/ParrotFlow/Config.swift
+++ b/Sources/ParrotFlow/Config.swift
@@ -85,9 +85,12 @@ struct Config: Decodable, Equatable {
         var namesBoth = false
         /// A `path:` that named a file this could not read, in words.
         var unreadable: String?
+        /// `tests: { path: heldout.yaml }` — the set `--eval` scores by default.
+        var tests: String?
 
         enum CodingKeys: String, CodingKey {
             case name, description, display, confirm, prompt, content, replace, command
+            case tests
             case timeout = "timeout_seconds"
         }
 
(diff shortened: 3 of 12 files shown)


znat:
## Valid, and fixed in 991a8a2

The warm-up ran `cases.first` — the first case *in the file*, not the first case *being scored* — so `--eval recorded --probe wanted` executed a case belonging to a probe the user had just excluded.

To this runner a transform is text in and text out, and an extra run is only an extra run. To a `command:` body it is not: that is someone's program, and nothing stops it writing a file or calling something. An input somebody explicitly filtered out is the last one to hand it.

`--probe` now selects once and the warm-up takes the first of what was selected, so it costs a second run of an input that was going to be scored anyway rather than a run of one nobody asked for.

## The case watches from the program's side

Rather than assert on `--eval`'s output, the transform is a script that appends every input it receives to a file — so the assertion is about what actually reached the program:

```
✓ --probe never hands an excluded input to the program
✓ and the warm-up runs a case that was going to be scored anyway
```

The set's *excluded* case is deliberately first, because that is what the old warm-up reached. **17/19 on the code this replaces, 19/19 with it.**

Two corrections I made to my own case while writing it, worth noting since both would have made it pass vacuously: `grep -c` exits 1 when the count is zero, so `$(grep -c … || echo 0)` produced `0\n0` and never matched; and with the wanted case first in the file, the old code never touched the excluded input at all.

All deterministic sets green, swiftlint clean.

znat:
## Valid, and fixed in 662188f — my previous fix was half a fix

This is not a new mistake; it is the one before it being incomplete, which is worth saying plainly.

`tests: { file: heldout.yaml }` recorded why the entry could not be used. The body was then assigned as a tuple into the *same variable*:

```swift
(body, source, unreadable) = readPrompt(promptPath)
```

So a `prompt: { path: … }` that read perfectly well overwrote the reason with `nil`. The entry was kept, and `--eval <name>` scored `cases.yaml` while the config named something else — exactly the silence the previous commit set out to remove, still there for any transform with a file-backed body.

The body's result goes through a local now, and the first reason an entry cannot be used is the one reported. A body that is fine has nothing to say about a key that is not.

The case pairs the two deliberately — a prompt file that reads *and* a `tests:` that does not — because either on its own already passed. **19/20 on the code it replaces, 20/20 with it.**

## One thing my own case taught me

The first version of it made three unrelated checks fail with `got 0`. The description I wrote was:

```yaml
description: a body that reads, and a tests: key that does not
```

A colon-space inside a plain scalar is not valid YAML, so the whole config stopped parsing and every later assertion silently measured nothing. It was only obvious because three checks failed at once rather than one — a single failure would have read as my fix not working, and I would have gone looking in the wrong place.

All eleven deterministic sets green, swiftlint clean.

znat:
## Valid, and fixed in f18a471 — and the second time is the interesting part

This is the same defect I fixed in the warm-up two commits ago, wearing a different hat. `badGold` sent every case's gold to the `resolve:` command before `--probe` had filtered anything.

`resolve:` is a `command:` like any other — someone's program, free to write a file or call something. Whether it reaches that program through the gold check or through the scoring loop makes no difference to the program.

**Fixing the same thing twice in two places is the signal that the shape was wrong.** The selection was being made *inside* `score`, where nothing else could see it, so every other place that runs a case had to remember to filter — and one of them didn't. It is made once in `run` now, before anything runs a case, and handed to both the gold check and the scoring. There is no second place left to forget.

That also lets `--probe` naming nothing be answered from the same list, so it is refused *before* the gold check rather than after the scoring — which stops a mistyped probe name from running a resolver at all.

## The case

Watches from the program's side again — a resolver that appends every gold it is handed — and asserts both halves, because a fix that resolved nothing would also pass the first one:

```
✓ the gold check never resolves an excluded case
✓ and it still checks the gold of the cases it will score
```

**21/22 on the code it replaces, 22/22 with it.**

## Still true

`code_identifiers` 68/75, `slack_mentions` 76/76. All deterministic sets green, swiftlint clean.

`check-eval.sh` has gone from 13 cases when this PR opened to 22, across six findings. Every one of them was real.

znat:
## Valid, and fixed in fffd7fd

`problems()` read every case in the file before `--probe` narrowed anything, so one case with no gold in it yet made **every** probe unscoreable — including the ones that were finished.

That is the wrong way round for what `--probe` is for. It is the flag you reach for while a set is being written: score the bucket you have done, go back to the one you are drafting. A case nobody is scoring cannot make a scored one wrong, and being unable to measure finished work because unfinished work sits in the same file is exactly the friction the flag exists to remove.

So validation now takes the selection — the third caller to be handed the same list, and the last one that was still deciding for itself.

Two faults stay whole-file, because neither can be selected around: a set with **no cases at all**, and an `intermediate:` with **no `resolve:`**, which is the thing that checks a gold against itself.

## A subset run says so

Otherwise `--eval x --probe finished` passing reads as a clean bill of health for the file, and the next full run is a surprise:

```
· 1 problem(s) in cases outside this probe, not checked here —
    run without --probe to see them
```

## The cases

Three, because the fix could go too far as easily as not far enough:

```
✓ a finished probe scores while another is still being written
✓ and it says the rest of the file was not checked
✓ while a full run still refuses the half-written set
```

**23/25 on the code it replaces, 25/25 with it.**

## Still true

`code_identifiers` 68/75, `slack_mentions` 76/76. All ten deterministic sets green, swiftlint clean.

Seven findings on this PR now, all valid. `check-eval.sh` has gone 13 → 25 cases. Three of the seven were the same defect — a selection being made in more than one place — which is why the selection is now made exactly once and passed to everything that reads a case.

PR #45, opened 2026-08-04
fix: a transform you can run is a transform you can ask for

znat:
Found by dictating into Slack and watching it go somewhere else.

```
inline router: "use slack handles." → slack
inline: slack rewrote the transcript
```

`slack` is the chat-tidying prompt. It tidied the sentence and left the name exactly as spoken, while `slack_mentions` — which scores 76/76 on its own set and turns that name into `@Alex MacLure` — sat unreachable.

## Why a missing tool is worse than it sounds

The catalogue the router chooses from was built from `config.prompts`, which is `transforms.compactMap(\.asPrompt)` — and `asPrompt` returns nil for a `command:` body. So a transform that became a script left the list without a word, and a `replace:` table was never in it at all.

That is not a tool going quiet. **A router shown nine of your ten tools does not report the tenth missing; it answers with the nearest of the nine.** From the outside a wrong route and a right route look identical: something happened, text came back. Which is why this survived being scored — every set in the repo measured the tools that were in the list.

## What changed

`Capability.transform` carries a `Config.Transform` rather than a `Config.Prompt`, and the catalogue is built from every transform that has a `description` to match spoken words against — **all three bodies**. What a transform is made of is not a reason you cannot ask for it out loud: a table costs nothing, a script costs a process start, and both answer to a description exactly as a prompt does.

An entry with no `description:` is not in the catalogue and `--check-config` names it apart, because there is nothing to match spoken words against. It still runs from a pipeline, where the name is written down rather than said.

One place now knows how to run a body — `perform` — so the two paths that reach a transform by voice, over a selection and inline in a dictation, cannot disagree about what running one means. It makes the same call `Pipeline.runTransform` does.

The spoken instruction still reaches a `prompt:` and nothing else. A script takes text on one side and gives text on the other; there is nowhere to put "in French" and no honest way to invent one.

## The cost, measured rather than assumed

A longer list is measurably harder to choose from — that is stated in `docs/pipelines.md` with numbers, and this makes the list longer. `tests/routing-cases.yaml` holds at **50/54**, unchanged, with four more capabilities in it.

`--check-config` prints what each capability is made of, because a program that answers your voice is a program run by saying a sentence, and this file says that out loud everywhere else it happens:

```
built-in  spelling — fix a name or word the speaker just spelled out letter by letter
prompt    slack — tidy dictated text into a chat message
program   slack_mentions — turn people's names into Slack mentions
table     backticks — wrap dotted paths in backticks, for chat
```

## Verification

| | |
|---|---|
| `check-routing.sh` | 50/54, the same as before |
| `--route "use slack handles"` | `slack_mentions` |
| `--route "turn the names into mentions"` | `slack_mentions` |
| `--route "wrap that in backticks"` | `backticks` |
| the eleven deterministic sets | unchanged |
| `swiftlint` | 0 errors |

`perform` makes the same `CommandRunner.run(in: transform.folder, …)` call the pipeline does, and `--eval slack_mentions` scores 76/76 through that path — so a script reached by voice runs where a script reached by a pipeline runs. The voice dispatch itself is the one thing no harness reaches; it has been installed and is being exercised by hand.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -917,10 +917,32 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         case .replace:
             return Replacements.applyExact(to: text, rules: transform.rules)
         case .command(let command):
-            // Nil is every way a program can fail, and it means keep the text.
-            return CommandRunner.run(
+            // Nil is every way a program can fail, and in a pipeline it means
+            // keep the text — a stage that fails must not cost you a sentence.
+            //
+            // Asked for by name it means something else. You said "use slack
+            // handles", the script did not run, and returning the text
+            // unchanged makes that indistinguishable from a script that ran
+            // and found nothing to do: the selection path then says "nothing
+            // to change" and the inline path says nothing at all. Both are
+            // describing a transform that worked. So this throws, and the
+            // failure paths both callers already have get to do their job.
+            guard let result = CommandRunner.run(
                 command, on: text, in: transform.folder, seconds: transform.timeout
-            ) ?? text
+            ) else { throw CommandDidNotRun(name: transform.name) }
+            return result
+        }
+    }
+
+    /// A `command:` transform that produced no rewrite, asked for by name.
+    ///
+    /// The log already carries which of the ways it went wrong — could not
+    /// start, exited non-zero, took too long, said nothing — and that is more
+    /// than fits on screen. What belongs on screen is that it did not run.
+    private struct CommandDidNotRun: LocalizedError {
+        let name: String
+        var errorDescription: String? {
+            "\(name) did not run — the log says what it said"
         }
     }
 
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -785,7 +785,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // being kept warm is Ollama's prompt cache, and a system prompt that
         // differs by one line is a cache miss and the 3.5s this exists to avoid.
         let system = Router.prompt(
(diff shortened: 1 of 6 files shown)


znat:
## Valid, and fixed in 64e017b

`perform` returned the text unchanged for every way a command can fail — could not start, exited non-zero, timed out, said nothing.

**In a pipeline that is right and deliberate.** A stage that fails must leave the transcript exactly as it arrived, because losing a stage costs a rewrite and losing a sentence costs the sentence.

**Asked for by name it is not.** You said "use slack handles", the script did not run, and unchanged text is indistinguishable from a script that ran and found nothing to do. So the selection path reported *"nothing to change"* — which describes a transform that worked — and the inline path reported nothing at all.

It throws now, and both callers already had a failure path waiting for it: the selection one flashes the reason, the inline one writes the dictated words **and** shows the reason for seven seconds. The transcript is still never lost, which is the rule that made the old behaviour tempting in the first place.

```
slack_mentions did not run — the log says what it said
```

Short on screen because the log already carries which of the four ways it went wrong, and that does not fit in a flash.

## The pipeline is untouched

`Pipeline.runTransform` does its own thing and still fails open. `check-pipeline.sh`'s *"a command transform keeps what it was given intact"* — a pipeline of `shout`, `broken` (exit 3) and `silent` (true) — is what says so, and it is still 27/27.

## On verifying this one

The two voice paths reach no harness in this repository, so this is verified by reading and by running the app, not by a case. Saying that plainly rather than listing green checks that do not cover it: everything below is a non-regression, not evidence that the new behaviour is right.

| | |
|---|---|
| `check-pipeline.sh` | 27/27 — fail-open unchanged |
| `check-routing.sh` | 50/54, the same as before |
| the other nine deterministic sets | unchanged |
| `swiftlint` | 0 errors |

Installed on my machine and being exercised by hand.

PR #48, opened 2026-08-05
feat: a stage can tell the next one what it just did

znat:
Every stage was `String -> String`, so the only thing one stage could tell the next was the sentence. A condition could ask "does the text still say *function*" and nothing else — which means a stage that had already done the work had no way to say so, and the stage below it re-derived the same judgement from the same words, or ran when it should not have.

`dotted` and `code_identifiers` are the pair that shows it: both fire in every terminal, both are about the same sentence, and neither can see the other. There was no way to write "only if nothing else took this".

Now there is:

```yaml
- transform: code_identifiers
- stage: transform
  transform: dotted
  when: code_identifiers.count == 0
```

## A stage returns what it contributed, not what it was given

A stage receives the whole scope and returns only its own part. That asymmetry is the design, not an economy: a stage has no way to spell "and everything else, unchanged", so it has no way to get that wrong. The pipeline is the only thing holding the accumulated scope and the only thing that could drop a namespace, and it never does.

Three rules follow, and each one closes a failure the obvious version invites:

**A stage writes under its own name.** The namespace comes from the pipeline, never from the stage. One stage reaching into another's facts is not guarded against — it is unspellable.

**Everything else carries through.** A stage contributing nothing erases nothing. Carrying is the loop's job rather than something each script has to remember, because a `sed` one-liner could never have echoed a namespace back, and a contract only some bodies can honour is not a contract.

**A name is a fact about the stage, not a claim about the text.** `code_identifiers.count` — how many names it converted — stays true after a later stage rewrites the sentence. A variable called `has_identifier` would quietly become a lie. Filing facts under the stage that produced them makes the first reading the natural one.

`ran`, `ok`, `changed` and `ms` are derived for every stage, whatever it is and whether or not it knows any of this exists. Derived, never claimed — a stage cannot forget to report `changed` and cannot report it about the wrong string. A skipped stage gets `ran` alone; inventing `ok` would let `grammar.ok` read true for a stage that never happened, so the way to ask is `grammar.ran && grammar.ok` and `&&` short-circuits for exactly that.

The built-ins publish real things too. `numbers.language` is the one worth naming: which grammar actually read the numbers, which is *not* the same answer as the pipeline's language, and which this app has computed and thrown away on the same line since numbers existed.

## Conditions that are not patterns are expressions

`/…/` still means a regular expression. Anything else is now an expression over the scope — a subset of [CEL](https://cel.dev), which is chosen for the specification rather than for any library. The evaluator is three hundred lines and any small grammar would have run; what a standard buys is that if this is ever rewritten in Python or Go, the configs people have already written keep meaning what they meant, because `cel-python` and `cel-go` implement the same operators.

```
paths          text, numbers.count, asr.confidence
literals       "a string", 12, 1.5, true, false
operators      &&  ||  !  ==  !=  <  <=  >  >=
methods        matches(re)  contains(s)  startsWith(s)  endsWith(s)
```

`!` is a real negation, which retires a footgun. Exclusion used to require an anchored negative lookahead, and `validate` had to refuse the unanchored form because `/(?!.*term)/` succeeds one character into "Terminal" and runs the stage everywhere it was written to stay out of. `!app.matches("term")` has no anchor to forget.

**An unknown name is an error, not false.** Silently false is how a condition stops working with nothing said, which is the failure this whole change exists to remove. Two of the three cases are caught at load, by `--check-config`, before a transcript reaches them:

- a name nothing defines — including `when: genre`, the old bare-word form
- a condition reading a stage declared **below** it, which no runtime error can catch in time: by then the transcript is already halfway through, and "has not run yet" and "reported nothing" are the same absence. Only the pipeline as a whole knows the order, which is the same argument that already refuses `fuzzy` before `replacements`.

### One thing is taken away

A bare word used to mean "this word appears, on word boundaries". It now parses as a variable. It was documented and used nowhere — not in the default config, not in an example, not in a fixture — and keeping it would have meant two condition languages in one key forever. `--check-config` refuses it by name and prints what to write instead:

```
numbers: when "genre" reads "genre", which is not a variable.
If you meant the word "genre" in the text, write it as a pattern: /genre/
```

## `returns: json`, declared and not sniffed

A `command:` gets the same channel by declaring it. In: `{text, ctx}`. Out: `{text?, vars?}` — both keys optional, because a stage that returns no `text` is saying it looked and left the sentence alone, which is different from echoing the input back and is not a failure.

Declared rather than detected, and the obvious alternative is why. "If stdout parses as an object with a `text` key, it is structured" breaks on exactly the transcript most likely to arrive in a folder that has a `code_identifiers` stage in it: one that *is* a JSON object. It would also make a script's protocol depend on its output, so a stage would change contract on one sentence in a thousand and nowhere else.

Off by default, and off is the whole of the old contract — stdout is the transcript, `sed` is still a transform, and nothing already written has to change. Scripts that never opt in still get the derived variables, so a script from before any of this is still something a later stage can gate on.

The app sets `PARROTFLOW_PROTOCOL=json` when — and only when — the transform declares it, so `returns:` stays the single declaration and a flag in the `command:` line cannot disagree with the key above it. It also keeps every script runnable by hand, which matters because every harness in `scripts/` pipes plain text.

## The default config is deliberately not switched over

`code_identifiers` still speaks the plain protocol on a new install.

A transform folder is written on first launch and **never overwritten**. Shipping `returns: json` in the default config would leave every existing install running the old `code_identifiers.py`, which knows nothing about the envelope, printing bare text into an app that would no longer read it — and that stage would silently stop doing anything. `config.example.yaml` shows the gating as a comment, and `authoring.md` names the hazard for anyone upgrading their own transform: edit the script first, then the config.

## Testing

`scripts/check-pipeline.sh` goes from 27 cases to **48**, all passing.

Twenty-one are on the variable machinery, scored against a script that publishes whatever its flags say rather than against a transform whose job is interesting — a failure is then about the carrying and never about whether some transform changed its mind about a word.

The override case is the one worth reading. A transform has one command, so two runs of it publish identical variables unless the script looks at what it published last time. So `counter` bumps its own count and labels itself only on the run that found nothing, and both halves of the rule fall out of one pipeline: `count` is republished and must end at 2, `label` is never mentioned again and must survive.

`--echo` is the half that cannot be faked in Swift — it reports back what the *script* found in its context, so a pass means the scope crossed a pipe rather than merely surviving in memory. `vars-shipped.yaml` does the same through the real `code_identifiers`, with `dotted` gated on what it published.

Five fixtures must not load at all, including the ordering case above.

**No regression on what this touched.** `code_identifiers` scores 38/43 change and 30/32 keep — identical before and after, measured by stashing the change and re-running. `check-numbers` 97/97, `check-dotted` 54/54, `check-replacements` 23/23, `check-eval` 25/25, `check-split`, `check-wake`, `check-transform-folders`, `check-seeded-transform`, `check-default-config` all unchanged.

`check-span` and `check-inplace` sit at 0/8 on this branch — and at 0/8 on a clean tree at `addf91b`, because they need the in-place substitution work that is not landed yet. Checked by stashing everything and re-running rather than assumed. `check-grammar` is 16/17 here and 16/17 at baseline.

## Also

`Replacements.exact` and `Numbers.read` are new overloads that report a count and the winning grammar; `applyExact` and `apply` are unchanged and still what every other caller uses. `Trace.recordStage` gained a `vars` column, so `trace.jsonl` answers "which transcripts did this actually fire on" with a `jq` query instead of a regex over English. The ASR numbers reach the pipeline from `Transcriber` rather than being read back off `Trace`, because the collector is only bound when someone asked for a trace and a condition must not work on the runs you are watching and stop on the runs you are not.

`--pipeline … --vars` prints the whole scope, and a skipped stage now names the values that decided it:

```
⊘ transform — skipped, when code_identifiers.count == 0 did not match (code_identifiers.count = 1)
```

## Not in this PR

The six transforms that funnel through `examples/transforms/_openai/openai_call.py` do not publish variables yet. That file is part of unlanded work in the working tree, and the `PARROTFLOW_PROTOCOL` branch it needs belongs with that change rather than with this one. Everything here stands without it.

https://claude.ai/code/session_01JT3WG3JPW2i7bLE9fynG5u

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,12 @@ rewrite a sentence and 8/8 asked only to identify which words matter, with a
 script doing the rest. Split the job before reaching for a bigger prompt.
 
 **A stage that costs a model call needs a condition.** `when:`, `unless:` or
-`app:`, so it is skipped on the transcripts that never needed it.
+`app:`, so it is skipped on the transcripts that never needed it. A condition
+between slashes is a regular expression; anything else is an expression over
+what the stages above published — `code_identifiers.count == 0`,
+`asr.confidence < 0.7`, `!grammar.changed`. `--pipeline <file> "<text>" --vars`
+prints the whole scope, and a skipped stage names the values that decided it.
+See [docs/pipelines.md](docs/pipelines.md#variables).
 
 **Say out loud what executes.** A `command:` transform makes `config.yaml` run
 a program. If you add one, tell the person whose machine it runs on, in the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,2 +1,21 @@
 See [AGENTS.md](AGENTS.md) — the entry point for any agent working in this
 repository, and the map to the documentation in [docs/](docs/README.md).
+
+## How to write
+
+Write in simplified technical English. This applies to **every** natural
+language output: chat replies, explanations, commit messages, pull request
+descriptions, issues, review comments, and documentation.
+
+- Short sentences. One idea per sentence.
+- Plain words. Say "use", not "leverage". Say "so", not "which is why".
+- Explain things simply. Assume the reader is smart but in a hurry.
+- No long sub-clauses, no em-dash chains, no rhetorical build-up.
+- Lead with the point. Then give the reason.
+- Keep it concrete: names, numbers, file paths.
+
+This is about the prose, not the work. Be just as precise and just as thorough.
+Say the same things — say them in fewer, simpler words.
+
+Code comments in this repository follow the surrounding style, which explains
+*why* a thing is the way it is. Keep that habit, but keep the sentences short.
--- a/Sources/ParrotFlow/CommandRunner.swift
+++ b/Sources/ParrotFlow/CommandRunner.swift
@@ -43,6 +43,85 @@ enum CommandRunner {
     /// that already runs on every transcript.
     static let timeout: TimeInterval = 2
 
+    /// What a command handed back.
+    ///
+    /// `text` is optional and that is the whole difference between the two
+    /// protocols. A bare command's stdout *is* the text, so it always has one. A
+    /// `returns: json` command may return only variables — "I looked, I found
+    /// three, I changed nothing" — and an absent `text` says that in a way an
+    /// echoed copy of the input cannot.
+    struct Output: Equatable {
+        var text: String?
+        var vars: [String: Scope.Value] = [:]
+    }
+
+    /// What a structured command is told about the transcript besides the
+    /// transcript.
+    ///
+    /// Read-only from the script's side, and deliberately not the same shape as
+    /// what comes back: a command receives the whole accumulated scope and
+    /// returns only its own contribution. It has no way to spell "and everything
+    /// else, unchanged", so it has no way to drop it. Carrying is the pipeline's
+    /// job — a `sed` one-liner could never have echoed a namespace, and a
+    /// contract only some bodies can honour is not one.
+    struct Context: Encodable {
+        let scope: Scope
+        let language: String
+
+        /// The bare names go at the top and the namespaced ones nest under
+        /// `vars`, so a script reads `ctx["vars"]["numbers"]["count"]` rather
+        /// than splitting a dotted string itself. The flat storage inside
+        /// `Scope` is an implementation detail of the evaluator, and pushing it
+        /// through the interface would make every script a parser.
+        func encode(to encoder: Encoder) throws {
+            var container = encoder.container(keyedBy: Key.self)
+            var nested: [String: [String: Scope.Value]] = [:]
+            for path in scope.paths {
+                guard let value = scope[path] else { continue }
+                guard let dot = path.firstIndex(of: ".") else {
+                    try container.encode(value, forKey: Key(path))
+                    continue
+                }
+                let namespace = String(path[path.startIndex..<dot])
+                let name = String(path[path.index(after: dot)...])
+                nested[namespace, default: [:]][name] = value
+            }
+            try container.encode(language, forKey: Key("language"))
+            try container.encode(nested, forKey: Key("vars"))
+        }
+
+        struct Key: CodingKey {
+            let stringValue: String
+            var intValue: Int? { nil }
+            init(_ value: String) { stringValue = value }
+            init?(stringValue value: String) { stringValue = value }
+            init?(intValue: Int) { return nil }
+        }
+    }
+
+    /// What goes in on stdin when a transform declares `returns: json`.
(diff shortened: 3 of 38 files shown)


znat:
Thanks — two of the three were real. Both are fixed in 2360485. The third is not right, and I explain why below.

## 1. Model conversions reported zero — valid, fixed

Correct, and it broke the exact condition this PR is built around.

`convert` counts names as the rules take them. When the rules decline and the model takes the name instead, `converted` stayed empty. So the stage published `count: 0` on a sentence it had just rewritten. A stage below it, told to stand down when the count is zero, would have run anyway.

`place` now reports what it converted, the same way `convert` does:

```python
out = place(ask(model, text), text, converted)
```

Checked directly:

```
text : call it max_retries
count: 1 -> ['max retries']
```

`code_identifiers` still scores 38/43 change and 30/32 keep.

## 2. Wrong language in the context — valid, and worse than reported

Correct. It was also a duplicate key, not just a wrong value.

The encoder wrote every bare scope name into the JSON, including the detected `language`. Then it wrote the parameter under the same key. The parameter was `languages.first`, so the wrong value overwrote the right one.

The fix removes the parameter. The context now takes everything from the scope and nothing beside it. One source for the field instead of two. `runCollectingScope` also seeds the detected language when the caller did not, so no path can leave it absent.

Two new cases in `vars-language.yaml` score this from the far side of the pipe — the script reports back what it was handed, so the test measures what arrived, not what Swift meant to send.

## 3. `--eval` runs JSON transforms as plain text — not right

`through` is the scoring path, and it runs the transform through `Pipeline.run`, which honours `returns: json`.

`piped` does use the plain path. But it is only used for `control:` and `intermediate:` helpers from the case file. Those are separate commands. Run bare, they get no `PARROTFLOW_PROTOCOL`, take the plain branch, and return plain text. That is consistent. This case set declares neither, so `piped` is never reached.

Measured, with `returns: json` set on the transform:

```
change   38/43 = 88%
keep     30/32 = 94%
overall  68/75 = 91%
```

Same as without it, and no `declares returns: json but printed …` in the log.

## And one you did not catch

Reviewing your points, I found four of my own cases were fake.

They had been appended after the `refused:` section, so they were read as refusals instead of cases. A refusal matches on a substring. An empty substring matches anything. So all four passed while never running — including both cases for the shipped transform.

They are back in `cases:`. A refusal with no `expect_problem` is now an error, and a missing refusal fixture is now an error too. A test that cannot fail is worse than no test.

Both fixes above are covered by cases that fail without them. I checked by putting each bug back and watching the case go red:

```
✗ a structured command is told the detected language  [vars-language]
✗ and the configured order does not decide it  [vars-language]
48/50
```

Then restored the fix: 50/50.

PR #50, opened 2026-08-06
feat: a prompt can see what is on your screen

znat:
Dictating into a terminal, the sentence is usually about what is already there — a file path, a function name, an error three lines up. The pipeline could not see any of it.

Two parts, shipped together because the second is what makes the first useful.

## 1. `context`, a new stage

Reads the screen behind the field and publishes it.

| variable | |
|---|---|
| `context.text` | the screen, minus the input box, last 2000 chars on a row boundary |
| `context.chars`, `.lines`, `.truncated` | what was read |
| `context.ok`, `.declined` | whether it read, and why not |

It never touches the transcript. That is by construction rather than by outcome, so `context.changed` is false on every run and means it — a stage that could put the screen into the transcript could paste your terminal into a chat message.

**It reads when the hotkey goes down, not when the stage runs.** The press is the last moment the pane is known: by then there has been a transcription and possibly a model call, and focus may be elsewhere. Reading late answers *what is on screen now* when the question is *what were you looking at when you decided to say this*.

Measured over 17 dictations: the two readings agreed 15 times, and both differences were under 35 characters of spinner and token counter. So the earlier one loses nothing and is the one that is certain. The read costs ~1ms on a small pane and 36–39ms on a long scrollback, so it runs off the main thread, and only when some pipeline names the stage.

**Terminals only.** A terminal's accessibility value *is* its visible screen, so the context is one call. A Slack composer publishes its own contents and nothing above it — the messages would need a walk of the window's children, hundreds of round trips per app. Worth building, not the same feature, and one stage meaning "cheap" in one app and "expensive" in the next is not one anybody can budget for. Everything else declines out loud with a reason.

**Not in the default pipeline.** It reads the screen. Turning that on for everyone who never wrote a `pipelines:` block would be a silent change to what the app looks at.

## 2. A prompt can read the scope

```yaml
prompt: |
  Rewrite the dictation as a clear instruction.

  This is on the speaker's screen right now:
  {{context.text}}

  Use it to spell names and paths. Never quote it back.
```

`{{context.text}}`, `{{numbers.count}}`, `{{language}}` — the same names `when:` reads, so there is one vocabulary rather than two. `{{instruction}}` stops being special-cased and becomes an ordinary lookup, which is also how it becomes visible that in a pipeline it was always empty.

**A name with nothing behind it takes its paragraph with it.** Most variables are absent most of the time; in Slack that middle paragraph is simply not there, heading included. Substituting an empty string would leave a heading over a blank, which tells a small model there is a screen and then shows it none.

The rule cannot save a prompt whose *every* paragraph holds a placeholder — there none is dropped, because an empty system message is not a smaller instruction. Two cases pin that down and the docs say to give a placeholder its own paragraph.

## Testing

```
scripts/check-compose.sh    21/21   new — offline, no model
scripts/check-pipeline.sh   54/54   +4
scripts/check-numbers.sh    97/97
scripts/check-dotted.sh     54/54
scripts/check-split.sh      14/14
scripts/check-replacements.sh 23/23
scripts/check-transform-folders.sh 12/12
scripts/check-default-config.sh ✓
```

`--compose "<template>" name=value …` renders a template with no model involved. What the model then does with the prompt is the other half and has no exact answer.

The capture itself is not scored and deliberately not faked: `--pipeline` has no hotkey, so there is never a capture to publish, and behind that TCC credits a terminal-run binary to the terminal. A fixture stubbing the screen would be scoring the stub. `--peek` shows a live capture.

## Two things to know before turning it on

- While the stage is enabled, `~/Library/Logs/ParrotFlow.log` holds what was on screen when you dictated.
- `{{context.text}}` in a prompt is the moment that screen travels to whatever `llm.endpoint` points at. The stage on its own never sends anything anywhere.

## Not in this PR

Where the text lands is unchanged. Dictate into one pane, switch before the transcript is ready, and the ⌘V still goes to the pane you switched to. The fix is to put the transcript on the clipboard rather than paste it somewhere you did not aim, and it belongs next to the paste.

https://claude.ai/code/session_01MwuvcJpErG3tV8e2itb3ES

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -431,6 +431,25 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             Log.write(String(format: "selection snapshot was slow: %.2fs", elapsed))
         }
 
+        // The screen as it was when you started talking — which is the screen
+        // the sentence is about. Taken here rather than in the pipeline because
+        // this is the last moment the pane is *known*: by the time the stage
+        // runs there has been a transcription and possibly a model call, and
+        // focus may be in another pane entirely.
+        //
+        // After the snapshot and off the main thread, because this reads an
+        // accessibility value that has been observed at 237k characters and the
+        // one thing this handler may not do is delay recording. Gated on the
+        // stage being configured, so a config that never asked for context pays
+        // nothing for the question.
+        if Context.isConfigured(in: config) {
+            let app = front?.app
+            let element = focusAtPress?.element
+            DispatchQueue.global(qos: .userInitiated).async {
+                Context.capturePress(app: app, element: element)
+            }
+        }
+
         switch config.hotkey.mode {
         case .toggle:
             toggleRecording()
--- a/Sources/ParrotFlow/ComposeCommand.swift
+++ b/Sources/ParrotFlow/ComposeCommand.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+/// `--compose "<template>" [name=value ...]` — prints a prompt with its
+/// placeholders filled, and applies nothing.
+///
+/// The whole point is that no model is involved. What a prompt *says* after the
+/// scope is folded into it is decided before any model sees it, and it is the
+/// half that can be scored exactly — which is what scripts/check-compose.sh
+/// does. `--prompt` exercises the other half and needs Ollama running.
+///
+/// Values are typed by how they read, because that is what a script writing a
+/// case file can express: `3` is an int, `true` is a bool, everything else is a
+/// string. A prompt renders all four the same way, so the distinction only shows
+/// up if a case ever needs it to.
+enum ComposeCommand {
+
+    static func run(template: String, assignments: [String]) -> Int32 {
+        var scope = Scope()
+        for assignment in assignments {
+            guard let split = assignment.firstIndex(of: "=") else {
+                print("✗ expected name=value, got \"\(assignment)\"")
+                return 2
+            }
+            let name = String(assignment[assignment.startIndex..<split])
+            let raw = String(assignment[assignment.index(after: split)...])
+            scope.set(name, value(of: expanded(raw)))
+        }
+        print(Template.fill(expanded(template), from: scope), terminator: "")
+        return 0
+    }
+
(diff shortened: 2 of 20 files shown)


znat:
Both review findings were real. Fixed in 392e404.

**Oversized row bypassed the limit** — confirmed. `tail` preferred a row boundary and, for the first row it kept, *required* one: the `!kept.isEmpty` guard existed to stop the result being empty and had no upper bound. A row longer than the whole budget came through whole.

Now the boundary is preferred rather than required. Where there is none inside the row it is cut from the front, keeping the end, like the function around it.

```
$PF --context-test 'ABCDEFGHIJKLMNOPQRSTUVWXYZ\nprompt$' 10
before:  ABCDEFGHIJKLMNOPQRSTUVWXYZ     (26 chars against a limit of 10)
after:   QRSTUVWXYZ
```

**Capture could cross dictations** — real, though it needs two hotkey presses inside one AX read. At the 1–39ms a healthy read takes that is implausible; against a wedged app the read runs to its messaging timeout, which is the case that makes it reachable. A generation counter taken at the start and checked before storing makes it unrepresentable — an overtaken read logs and drops rather than putting the older screen back.

**New case set.** `--context-test "<screen>" [limit]` runs the two string steps of the stage — cut the input box off, keep the tail — with no accessibility involved. `scripts/check-context.sh` scores 7 cases: the TUI box, the bare shell, blank padding, the tail, the under-limit flag, and the oversized row both alone and with a row that fits below it.

Reverting the truncation fix fails that case, which is the point of having it:

```
  want:   truncated | QRSTUVWXYZ
  got:    truncated | ABCDEFGHIJKLMNOPQRSTUVWXYZ
  6/7
```

All suites on the branch:

```
check-context 7/7    check-compose 21/21   check-pipeline 54/54
check-numbers 97/97  check-dotted 54/54    check-split 14/14
check-replacements 23/23   check-transform-folders 12/12   check-default-config ✓
```

One note on the summary's framing: the shared slot and the truncation path are ordinary correctness bugs in a stage that reads your own screen into your own log and your own local model. Calling them privacy-boundary defects overstates it — the value never left the machine unless `llm.endpoint` is remote, which the docs already call out. The findings were still worth having and both are fixed.

PR #51, opened 2026-08-06
feat: a dictation you can stop, and one that will not land in the wrong window

znat:
Two ways a dictation could end up somewhere you did not want it.

## Escape stops a dictation

Press ⎋ while recording, or while it is transcribing. The dictation ends and nothing is written.

**It works with the hotkey still held.** That is the point of it. In push-to-talk the key is down for the whole sentence and letting go is what commits the recording, so a cancel you had to release first would be no cancel at all. Stopping the recorder means the later release finds nothing recording and does nothing.

Cancelling during transcription cannot make the decoder stop sooner — the call is not interruptible. What it guarantees is that the result goes nowhere when it lands, and says so in the log rather than disappearing.

The clip stays in your recordings folder. It is where you go to hear what the app heard, and the moment you cancelled is the moment you most want it. It is simply never transcribed.

The listener is installed when recording starts and removed when the dictation is over, so an idle app is not watching the keyboard. It observes rather than consumes, so the app you are typing into still sees the ⎋ — swallowing it would need an event tap over every keystroke on the machine, and stopping a dictation into a TUI usually means you want the TUI to see it too.

Scope: recording and transcription. A voice command or an inline transform running after the transcript has arrived is not covered.

## A transcript that lost its window is copied, not pasted

`TextInserter` posts ⌘V into whatever is frontmost, and a transcript arrives seconds after you stop talking — a decoder, then whatever the pipeline runs. Dictate an instruction into one terminal pane, switch to the next while it works, and the sentence landed in the wrong session.

Nothing in this app has ever pinned a pane. Every `owner.activate()` here activates an *application*.

So the focused element is compared against the one that was there at the press, and a transcript with no home goes to the clipboard with a notice.

It does not try to steer focus back. That would mean setting `kAXFocused` and trusting an app to honour it, which is a guess. The clipboard is not a guess.

**On the risk of false positives** — guard too eagerly and every dictation ends up copied, which would make this worse than the bug. The comparison was measured before it was relied on: the press-time experiment behind the `context` stage (#50) ran this same `CFEqual` across 17 real dictations and the element was equal every time.

`aimedAt` is passed at all nine call sites rather than defaulted, so a path added later has to say whether it has an aim.

## Testing

```
check-context 7/7     check-compose 21/21   check-pipeline 54/54
check-split 14/14     check-numbers 97/97   check-default-config ✓
```

**Neither change has a case set, and that is worth being explicit about.** Both live in the hotkey and paste paths. The CLI harness has no hotkey and nothing focused, so it cannot reach either — the same reason the `context` capture is scored by `--peek` rather than by a fixture. What can be checked is that nothing else moved, and nothing did.

Worth exercising by hand:

- hold the hotkey, press ⎋ mid-sentence, release — nothing is written, and the next press still records
- press ⎋ while "Transcribing…" is up — nothing is written
- dictate into one Ghostty pane, switch panes before it lands — it should copy and say so
- dictate and stay put — it should paste exactly as before

https://claude.ai/code/session_01MwuvcJpErG3tV8e2itb3ES

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -92,6 +92,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// Running between the hotkey coming up and the recording actually stopping
     /// — see `stopRecordingAfterTail`.
     private var releaseTail: Timer?
+    /// Every transcription run at or below this was cancelled with Escape and
+    /// must not be written anywhere. A high-water mark rather than a set: runs
+    /// only ever increase, so "this one and everything still in flight behind
+    /// it" is the whole of what cancelling means.
+    private var cancelledThroughRun = 0
+    private var escapeMonitors: [Any] = []
     private var keepWarmTimer: Timer?
     private var keepWarmInFlight = false
     private var hotkeyError: String?
@@ -580,6 +586,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             return
         }
 
+        watchForEscape()
+
         if config.feedback.sound { NSSound(named: "Tink")?.play() }
         if config.feedback.overlay {
             overlay.model.elapsed = 0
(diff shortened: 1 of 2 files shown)


znat:
Both findings were real. Fixed in the commit above.

**Older runs removed newer monitors** — confirmed, and it is more reachable than it looks. The app deliberately supports overlapping dictations: push-to-talk does not wait for the previous transcript, which is what `transcriptionRun` exists to survive. So the sequence is ordinary — dictate, hold the key again while a prompt stage is still running, and the second recording shares the monitor pair the first one installed. The first run finishing then took them away, and Escape stopped working for the recording still going. It would only ever show up on a second press.

`runsInFlight` now counts outstanding transcriptions, and teardown is conditional:

```swift
private func stopWatchingForEscapeIfIdle() {
    guard !recorder.isRecording, runsInFlight <= 0 else { return }
    stopWatchingForEscape()
}
```

Cancelling is unchanged — `cancelledThroughRun` already means "this run and everything still behind it", so one Escape stops both.

**Failed focus lookup permitted a paste** — also real. `SelectionReader.focusedElement()` returns nil on a 0.25s messaging timeout against a busy app, and `let now = …` in the condition meant that fell straight through to an untargeted ⌘V. That is the guess the whole check exists to avoid.

Nil now counts as moved. Being wrong that way puts the transcript on the clipboard with a notice, which is recoverable. Being wrong the other way puts a sentence in someone else's window.

One thing I did not take from the finding: the check now tests Accessibility *first*. Without the grant, `focusedElement()` returns nil for that reason rather than because anything moved, and `TextInserter` already returns `.clipboardOnly` with "grant Accessibility to auto-paste" — a better message than "Focus moved". Preempting it would have made a permissions problem look like a focus problem.

```
check-context 7/7    check-compose 21/21   check-pipeline 54/54
check-split 14/14    check-numbers 97/97   check-transform-folders 12/12
check-default-config ✓
```

Still no case set for either path, for the reason in the PR description — the harness has no hotkey and nothing focused. The overlap case is worth exercising by hand: dictate, press the hotkey again while "Transcribing…" is still up, then press ⎋ and check the second recording actually stops.

PR #62, opened 2026-08-08
test: the measurement harness, with the bugs the review found taken out

znat:
PR 2 of `docs/proposals/vocabulary-v2.md`. Test-only: nothing under `Sources/`
changes.

## What was ported

From the frozen `feat/vocabulary-skills-only`, read with `git show`, never
committed to:

| File | |
|---|---|
| `scripts/menu-recall.py` | did the true sentence reach the judge, did the judge take it |
| `scripts/mine-menu-cases.py` | builds the labelling sheet out of the archive |
| `scripts/tune-judge.py` | scores a judge prompt against the cached menus |
| `scripts/before-after.py` | what a clip produced then, what it produces now |
| `scripts/rerank-judge.py` | the reranker spike and the two-stage harness |
| `tests/menu-cases.yaml` | 40 clips, 37 labelled |
| `tests/judge-menus.json` | 33 cached menus, 28 reachable |
| `tests/judge-cases.yaml` | the 2026-08-07 relabels (the four "spells" cases are declines) |

One file the plan does not list came with them:
`examples/transforms/verify_names/menu.md`. It is the judge prompt that
`tune-judge.py` and `rerank-judge.py --stage2` score. Without it neither
harness runs. PR 3 renames it to `verify_names.md` and installs it; nothing
installs `examples/` today.

Every F13 fix already in these scripts is kept: the built app before the
installed one, the first `trace.jsonl` entry per clip, the non-ASCII guard in
`ranking()`.

## The fixes

**F7 — `--stage2` shortlist order.** The shortlist is now presented in the
app's own menu order, decoder's reading first. The reranker chooses *which*
readings survive, not the order they are read in.

**F7 — `--stage2` score block.** The block is trimmed to lines that are still
about this shortlist. A line survives two tests: it has to be a real
measurement (not the 0.00 sentinel, F6), and the shortlisted options have to
still disagree about it.

Ablated on the cached menus, mxbai-rerank-base-v2, framing `misheard`, top-3,
judged by `gemma4:e4b`:

| | picked |
|---|---|
| neither fix | 23/28 |
| order only | 24/28 |
| discriminating trim only | 23/28 |
| both, sentinel lines kept | 23/28 |
| **both, sentinel lines dropped** | **25/28** |

This differs from the ledger, which recorded the trim as worth +1 on its own.
It is not. The trim on its own can strand a sentinel line as the only survivor
— on `2026-08-07T17-05-42` it dropped the two real Supabase lines and kept
[...]

PR #66, opened 2026-08-08
feat: a name may claim two places in a sentence, not six

znat:
PR 6 of `docs/proposals/vocabulary-v2.md`. Names arrive split ("praise he"),
possessed ("Mirza's") and punctuated ("Versailles."), and every reading built
from the shorter span strands a word.

**Four of PR 6's five items were already on `main`.** PR 3's own `Do` list told
it to port "the acoustic search, the wider-span variants, and the possessive
carry" and "widest-first ranking, per-slot cap". It did. The three clips this
PR is supposed to fix already behaved before it started, and the numbers below
say so: recall and before/after do not move.

The fifth item did not land. Nothing capped what **one term** contributes to a
menu, and that is the cap the plan reserved as this PR's headroom knob.

## What was already there, and where

| Item | Where | Evidence |
| --- | --- | --- |
| span, span+1, span+2 × `term`, `term + 's`, each pair tested independently | `Vocabulary.widerSpans` | `"praise his" -> "Praisy" (span 0.45)` and `"praise his" -> "Praisy's" (span 0.59)` both offered on `17-39-03` |
| a trailing possessive carried through a substitution | `Vocabulary.inflected` | `"Mirza's" -> "Mirza" applied` and the transcript still reads `Mirza's` |
| trailing punctuation trimmed before replacing | `Vocabulary.trailingMarks`, plus the trim at the end of `acousticSpans` | `"Pressy." -> "Praisy" applied` → `Let's praise Praisy.` |
| widest span first, possessive as tiebreak | `VocabularyJudge.slots` | Identical to `menu.py` line 237 on the frozen branch |

Same as the prototype in every case, with one deliberate narrowing PR 3
recorded: a possessive is not offered at a sentence end.

## What this PR adds

**`max_per_term: 2`.** A name reaches a menu from four directions at once — a
`replacements` rule that already rewrote the text, the rescorer's own proposal,
the wider spans built around it, and the CTC spotter hearing it somewhere else
in the clip. `max_per_slot` bounds one place. Nothing bounded the total, so a
noisy name spends slots that belong to other names.

```yaml
- vocabulary: verify_names.md
  when: vocabulary.count > 0
  max_per_term: 2     # optional; places in one sentence about the same name
```

**Which two survive is a rank, not a score,** because the four sources share no
number: a rule has none at all, a wider span was never measured acoustically,
and the spotter scores the term without scoring the word the decoder wrote. So
`VocabularyJudge.Standing` — rule, then scored, then wide, then spotted — with
the earliest place breaking a tie. The cap is applied to finished slots, so a
group whose readings all collapse to the decoder's own cannot spend a term's
budget on its way to being dropped.

**The judge logs its slot count on every run.** `max_slots: 4` is a cliff: one
over and the whole menu is declined, which in a transcript looks exactly like a
clip nobody proposed anything for. It was logged only when it was already
fatal.

```
vocabulary judge: 6 slot(s) from 6 proposal(s) — "Matthieu" (Matthieu), "went to the" (Matthieu), "Vercel" (Vercel), "universal" (Vercel), "deployed on" (Claude), "Vercel" (Vercel)
```

**Two comments corrected.** `Vocabulary.spansPerTerm` said it capped "readings
one term may contribute to a menu"; it gates the spotter only, and never saw
the other three sources. `readings()` said the trim takes the wider spans
first; since the widest-first ranking landed it takes the narrowest.

## The cap's justification, and what it is not

Menu readability, and nothing else. `max_per_slot`'s doc comment now says so
out loud and names what it is *not* leaning on: the three-option shortlist that
scored the same as the full menu is F7's, F7 was the harness reordering the
shortlist, and PR #62's ablation gave the effect back to the ordering. The
menu was not resized on that number. `max_readings` now admits it was never
measured either — it was set to match `max_slots`.

`max_per_term: 2` says a name said twice in one sentence is ordinary and a
third mention is more often the spotter than the speaker. Swept rather than
chosen, one run each, everything else at the shipped settings:

| `max_per_term` | recall | picked |
| --- | --- | --- |
| 1 | 28/37 | 24/37 |
| **2** | **31/37** | **27/37** |
| 3 | 31/37 | 27/37 |
| 99 (off) | 31/37 | 27/37 |

2 is the tight end of the plateau, which is the end to take for a cap whose job
is headroom. **The spotter floor was not moved** — it is still −5.0, where
PR #65 left it.

## Slot counts, before and after

The risk this PR was handed: widening pushes slot counts past `max_slots`, the
judge declines everything, and recall drops. Measured on the 40 clips of
`tests/menu-cases.yaml`, per-clip majority of 3 runs:

| slots | before | after |
| --- | --- | --- |
| 0 | 5 | 5 |
| 1 | 17 | 17 |
| 2 | 14 | 14 |
| 3 | 3 | 3 |
| 4 | 0 | 0 |
| 5 | 0 | **1** |
| 6 | **1** | 0 |
| **over `max_slots: 4`** | **1** | **1** |

Nothing widened, so nothing moved except the one clip the cap acts on.

**`17-39-40` is that clip, and it is worth reading.** Six slots for three
terms:

```
"Matthieu"     (Matthieu)   rule — offers "Mathieu" back
"went to the"  (Matthieu)   spotter -4.98
"Vercel"       (Vercel)     rule — offers "Versailles" back   ← the answer is here
"universal"    (Vercel)     spotter -2.51
"deployed on"  (Claude)     spotter -4.65
"Vercel"       (Vercel)     rule
```

`max_per_term: 2` drops `"universal"`, which is the right one to drop and the
one the rank picks:

```
vocabulary judge: "universal" is a Vercel reading past max_per_term 2; not offered
vocabulary judge: 5 slot(s) from 6 proposal(s) — …
pipeline: vocabulary — 5 slots > 4; kept as decoded
```

Five is still over four, so the clip is still BROKEN — but it now comes out
right on 1 run in 3 instead of 0, which is the flip counted below. Its two
remaining spotter slots, `"went to the" → Matthieu` and `"deployed on" →
Claude`, are not names anybody said. That is an evidence problem, which is
PR 7's, not a span one.

## Gates

Built app, clean tree, stamp matching HEAD:

```
$ swift build -c release && scripts/build-app.sh
==> Build stamp: f7e7d1e
$ .build/ParrotFlowDev.app/Contents/MacOS/ParrotFlow --version
f7e7d1e
$ git rev-parse --short HEAD
f7e7d1e
```

Measured first on `6af93d6` and re-measured, whole, on `f7e7d1e` after
Greptile round 1. Both give the same four numbers and the same per-clip
verdicts. The two commits between them are `2ce6077` (renames an `Int`
parameter of `rewritten`) and `f7e7d1e` (puts the census words behind
`PARROTFLOW_JUDGE_DUMP`, and adds a refused fixture) — neither is on the
transcription path.

Scratch config dir built the way PR #63/#64/#65 document: the live
`config.yaml` and `vocabulary.yaml`, the pipeline entry replaced by
`- vocabulary: verify_names.md  when: vocabulary.count > 0` with `context` and
`fuzzy` moved below it, the prompt copied in beside it, `llm.model` set to
`gemma4:e4b`, and an empty scratch `voice/`. `~/.config/parrotflow-dev/` was
not touched and `make install` was not run. Other models were unloaded with
`keep_alive: 0`; Nathan's installed app reloaded its own `gemma4:e4b-mlx`
during the runs, as it did in PR #65, and the per-clip verdicts are identical
across baseline and after, so it moved nothing.

```
$ PARROTFLOW_CONFIG_DIR=<scratch> python3 scripts/menu-recall.py --runs 3
  recall  31/37   the true sentence was on the menu, majority of 3 runs
  picked  27/37   the judge chose it, majority of 3 runs
  0/37 clip(s) changed outcome between runs (F12a)
  (3 clip(s) unlabelled)

$ PARROTFLOW_CONFIG_DIR=<scratch> python3 scripts/before-after.py --runs 3
  fixed 17   broken 9   regressed 1   already right 10
  majority of 3 runs; 1 clip(s) changed outcome between runs (F12a)
    parrotflow-2026-08-07T17-39-40.wav  right on 1/3
```

Baseline measured by me on this machine today, on `824d402` — `main` plus the
slot-count log line, which cannot move a transcript:

| | baseline | this PR | gate |
| --- | --- | --- | --- |
| recall | 31/37 | 31/37 | ≥ 31/37 ✓ |
| picked | 27/37 | 27/37 | ≥ 27/37 ✓ |
| before/after | 17 / 9 / 1 | 17 / 9 / 1 | no worse ✓ |

**Per-clip diff: no clip changed verdict in either harness.**

```
menu-recall:   40 clips base, 40 clips after, 0 moved
before-after:  37 clips base, 37 clips after, 0 moved
```

**Flips.** `menu-recall.py` 0 in both. `before-after.py` 0 in the baseline, 1
here: `17-39-40`, right on 1 of 3 runs, verdict unchanged at BROKEN. That clip
sits on the `max_slots` boundary and the cap moved it from 6 slots to 5, so
replay noise now occasionally takes it under four. It is the only clip in the
set anywhere near the cliff.

The one REGRESSED row is `17-47-45`, unchanged since PR #63: `retry` versus
`Arexvy`, 0.46 apart, the audio prefers `retry` by 0.25 nats. A judge error,
not a mechanism one.

## The three named clips

**"praise he / praise his" — the wide reading is on the menu** (`17-39-03`):

```
vocabulary: "praise his" -> "Praisy"   also offered (span 0.45)
vocabulary: "praise his" -> "Praisy's" also offered (span 0.59)
vocabulary judge: 2 slot(s) from 9 proposal(s) — "praise" (Praisy/Praisy's), "praise his" (Praisy/Praisy's)
MENU C. Let's praise Praisy's work.
── transcript ─────────────────────────────
Let's praise Praisy's work.
```

Both readings of the wide span are offered, and each cleared the floor on its
own — 0.45 and 0.59 are on opposite sides of anything that would have tested
the possessive only where the plain term already passed.

**"Mirza's" — the possessive survives** (`17-39-19`):

```
vocabulary: "Mirza's" -> "Mirza" applied (raw -5.56 vs -6.39, bonus 2.88)
MENU A. Let's praise Mirza's work.
```

The rescorer replaces the whole token; `inflected` puts the `'s` back, so an
auto-applied substitution keeps the grammar the decoder had right.

**"Versailles." — the full stop survives.** Two clips, because the archive
holds the sentence-final case and the mid-sentence one separately.

`15-37-32`, a rendering with a full stop replaced by its term:

```
vocabulary: "Pressy." -> "Praisy" applied (raw -10.55 vs -11.71, bonus 2.88)
── transcript ─────────────────────────────
Let's praise Praisy.
```

`17-47-45`, the same thing reaching a menu rather than being written:

```
vocabulary judge: 3 slot(s) from 6 proposal(s) — "retry." (Arexvy.), "retry" (Arexvy), "crawl" (Redcrawl)
MENU E. Oh I get it. So yes, instead we want to have a Arexvy. And then if the retry fails, we want the crawl to fail.
```

The reading on the menu is `Arexvy.` and not `Arexvy`, so no option on the list
is a sentence nobody could have said.

And `16-04-42`, where `Versailles` is offered back on both sides of one
sentence:

```
MENU B. Mirza and Mira went to the movie. The movie was about the Versailles Castle. … being the plot on Vercel.
```

## Repository checks

All on `f7e7d1e`:

```
check-pipeline                  57/57   [exit 0]   (a refused fixture for max_per_term: 0)
check-replacements              23/23   [exit 0]
check-transform-folders         12/12   [exit 0]
check-default-config           ✓ a new install gets every key config.example.yaml documents  (28 keys, 9 transforms, 2 transform step(s))   [exit 0]
check-seeded-transform         ✓ transforms/code_identifiers/cases.yaml is the one in examples/  (489 lines)   [exit 0]
check-eval                      25/25   [exit 0]
check-numbers                   97/97   fr 53/53   en 26/26   auto 18/18   [exit 0]
check-dotted                    54/54   plus 2 known-unfixable   [exit 0]
check-split                     14/14   [exit 0]
check-wake                      24/24   plus 1 known-unfixed   [exit 0]
check-dates                     26/26 in 2s   1 known limit(s), not counted   [exit 0]
check-pinned-certificate       ✓ both install paths pin the same certificate  1fe06cb4b110d3f6…   [exit 0]
check-no-voice                  5/5   [exit 0]
check-vocabulary-config         61/61   [exit 0]
```

`check-span.sh`, `check-inplace.sh`, `check-grammar.sh`, `check-routing.sh` and
`check-spelling.sh` were not run: the first two need `make install`, a real
screen and Accessibility, and none of the five touches anything on this branch.

A bad cap is refused rather than run, in the shape `--check-config` already
uses for the other three:

```
$ ParrotFlow --check-config       # max_per_term: 0
✗ pipeline en: vocabulary: max_per_term is 0 — it has to be at least 1
```

## Review

Two rounds with `greptile-apps[bot]`.

| Round | Finding | Outcome |
| --- | --- | --- |
| 1 (P1, security) | The slot census wrote each slot's dictated substring and its terms to the persistent log on every run that produced a slot, including declined ones | Fixed in `f7e7d1e`: counts stay unconditional, words moved behind `PARROTFLOW_JUDGE_DUMP` |
| 1 (P2) | The cap's survivor ranking has no deterministic test | Half fixed: a refused fixture covers the validation contract, `check-pipeline.sh` 57/57. The ranking half is argued — there is no Swift test target, `--pipeline` and `--replace` have no audio so every part is `.rule` and `Standing` never has two values to order, and the input that would exercise it is a wav of the speaker's voice that `check-no-voice.sh` keeps out of the repository |
| 2 | 8 files reviewed, 0 comments added | — |

## Deviations

- **Four of five items were verification, not implementation.** Reported above
  rather than re-ported. Re-porting `widerSpans` from the frozen branch would
  have reverted PR 3's sentence-end guard, which is worth a case.
- **The cap counts places, not readings.** The plan says "cap readings per
  term at 2–3". Read literally that is `max_per_slot`, which already ships at
  3, and applying a reading cap of 2–3 *across* slots breaks the `Praisy`
  clips, where one term legitimately fills two places with three readings
  each. Two readings of one span — `Praisy` and `Praisy's` over "praise" — are
  one question, not two, and counting them separately spends the budget on the
  case the stage exists for.
- **A new stage param, not an env var,** which is what the plan's decisions
  table settles. The sweep above was run by editing `max_per_term:` in the
  scratch config, not by adding an override the harness would then own.
- **No new script.** The slot-count table came from a throwaway reader of the
  log line this PR adds. The log line is the durable part.
- **No live dictation.** Every number is a replay, `--runs 3`, with its flip
  count. Nothing here is HUMAN-gated.

### DIFF
--- a/Sources/ParrotFlow/Config.swift
+++ b/Sources/ParrotFlow/Config.swift
@@ -1256,6 +1256,7 @@ struct Config: Decodable, Equatable {
                 case maxSlots = "max_slots"
                 case maxReadings = "max_readings"
                 case maxPerSlot = "max_per_slot"
+                case maxPerTerm = "max_per_term"
             }
 
             init(from decoder: Decoder) throws {
@@ -1273,7 +1274,7 @@ struct Config: Decodable, Equatable {
                     prompt = judged
                     var caps = VocabularyJudge.Caps.standard
                     // Each optional and each on its own: a person raising the
-                    // menu ceiling should not have to restate the other two.
+                    // menu ceiling should not have to restate the rest.
                     if let slots = try c.decodeIfPresent(Int.self, forKey: .maxSlots) {
                         caps.slots = slots
                     }
@@ -1282,6 +1283,9 @@ struct Config: Decodable, Equatable {
                     }
                     if let perSlot = try c.decodeIfPresent(Int.self, forKey: .maxPerSlot) {
                         caps.perSlot = perSlot
+                    }
+                    if let perTerm = try c.decodeIfPresent(Int.self, forKey: .maxPerTerm) {
+                        caps.perTerm = perTerm
                     }
                     self.caps = caps
                     namesBoth = stage != nil || named != nil
--- a/Sources/ParrotFlow/Pipeline.swift
+++ b/Sources/ParrotFlow/Pipeline.swift
@@ -920,15 +920,24 @@ struct Pipeline: Equatable, Codable {
         ) + VocabularyJudge.ruleParts(rules, in: text, before: findings?.text)
 
         let slots = VocabularyJudge.slots(in: text, from: parts, caps: caps)
-        // Logged on every run, not only when the count is fatal. `max_slots` is
-        // the cliff this stage falls off — one slot over and the whole menu is
-        // declined — so the distance to it has to be measurable before a change
-        // that widens what fires, not inferred afterwards from the clips that
-        // broke.
-        Log.write("vocabulary judge: \(slots.count) slot(s) from \(parts.count) proposal(s)"
-            + (slots.isEmpty ? "" : " — " + slots.map {
+        // Two numbers on every run, not only when the count is fatal.
+        // `max_slots` is the cliff this stage falls off — one slot over and the
+        // whole menu is declined — so the distance to it has to be measurable
+        // before a change that widens what fires, not inferred afterwards from
+        // the clips that broke.
+        //
+        // Counts only. **Which words** is behind `PARROTFLOW_JUDGE_DUMP`, the
+        // switch that already means "write this dictation's menu down for a
+        // harness to read". The log is a plain file under `Library/Logs` and it
+        // outlives the dictation, so a diagnostic that is on for everybody
+        // spells names into it on runs where nothing was even offered.
+        var census = "vocabulary judge: \(slots.count) slot(s) from \(parts.count) proposal(s)"
+        if !slots.isEmpty, ProcessInfo.processInfo.environment["PARROTFLOW_JUDGE_DUMP"] != nil {
+            census += " — " + slots.map {
                 "\"\(text[$0.range])\" (\($0.terms.joined(separator: "/")))"
-            }.joined(separator: ", ")))
+            }.joined(separator: ", ")
+        }
+        Log.write(census)
         guard !slots.isEmpty else {
             return StageResult(text: text, vars: ["asked": .int(0), "slots": .int(0)])
         }
--- a/Sources/ParrotFlow/Pipeline.swift
+++ b/Sources/ParrotFlow/Pipeline.swift
@@ -925,7 +925,10 @@ struct Pipeline: Equatable, Codable {
         // declined — so the distance to it has to be measurable before a change
         // that widens what fires, not inferred afterwards from the clips that
         // broke.
-        Log.write("vocabulary judge: \(slots.count) slot(s) from \(parts.count) proposal(s)")
+        Log.write("vocabulary judge: \(slots.count) slot(s) from \(parts.count) proposal(s)"
+            + (slots.isEmpty ? "" : " — " + slots.map {
+                "\"\(text[$0.range])\" (\($0.terms.joined(separator: "/")))"
+            }.joined(separator: ", ")))
         guard !slots.isEmpty else {
             return StageResult(text: text, vars: ["asked": .int(0), "slots": .int(0)])
         }
--- a/Sources/ParrotFlow/Pipeline.swift
+++ b/Sources/ParrotFlow/Pipeline.swift
@@ -920,6 +920,12 @@ struct Pipeline: Equatable, Codable {
         ) + VocabularyJudge.ruleParts(rules, in: text, before: findings?.text)
 
         let slots = VocabularyJudge.slots(in: text, from: parts, caps: caps)
+        // Logged on every run, not only when the count is fatal. `max_slots` is
+        // the cliff this stage falls off — one slot over and the whole menu is
+        // declined — so the distance to it has to be measurable before a change
+        // that widens what fires, not inferred afterwards from the clips that
+        // broke.
+        Log.write("vocabulary judge: \(slots.count) slot(s) from \(parts.count) proposal(s)")
         guard !slots.isEmpty else {
             return StageResult(text: text, vars: ["asked": .int(0), "slots": .int(0)])
         }
--- a/Sources/ParrotFlow/Vocabulary.swift
+++ b/Sources/ParrotFlow/Vocabulary.swift
@@ -375,9 +375,17 @@ actor Vocabulary {
         "will", "would", "can", "could", "should", "my", "your", "our", "their",
     ]
 
-    /// How many readings one term may contribute to a menu. The cap is the
-    /// point: recall wants every plausible span offered, and a person reading
-    /// a lettered list wants a list they can read.
+    /// How many **spotter** spans one term may contribute. Not the menu.
+    ///
+    /// It read "how many readings one term may contribute to a menu", which is
+    /// not what it does and not where it is used. It gates `acousticSpans`
+    /// only, so it never saw the rescorer's own proposals, the wider spans, or
+    /// a `replacements` rule — and a term reaches a menu from all four. The
+    /// menu-level cap is `VocabularyJudge.Caps.perTerm`, which is applied where
+    /// the four meet.
+    ///
+    /// Kept where it is because it is cheaper here: a detection refused now
+    /// never becomes a proposal, and a 19-second clip yields ninety of them.
     static var spansPerTerm: Int {
         ProcessInfo.processInfo.environment["PARROTFLOW_SPANS_PER_TERM"]
             .flatMap(Int.init) ?? 2
--- a/Sources/ParrotFlow/VocabularyJudge.swift
+++ b/Sources/ParrotFlow/VocabularyJudge.swift
@@ -352,7 +352,7 @@ enum VocabularyJudge {
     /// also wrote the term. Guessing there is how a correct word gets a wrong
     /// spelling put first on the menu.
     static func rewritten(
-        _ heard: String, _ term: String, in before: String, became standing: Int
+        _ heard: String, _ term: String, in before: String, became stands: Int
     ) -> [Int]? {
         var marks: [(at: String.Index, rule: Bool)] =
             Vocabulary.spans(of: heard, in: before, ignoringCase: true)
@@ -360,7 +360,7 @@ enum VocabularyJudge {
             + Vocabulary.spans(of: term, in: before, ignoringCase: true)
                 .map { ($0.lowerBound, false) }
         marks.sort { $0.at < $1.at }
-        guard marks.count == standing else { return nil }
+        guard marks.count == stands else { return nil }
         return marks.indices.filter { marks[$0].rule }
     }
 
--- a/Sources/ParrotFlow/VocabularyJudge.swift
+++ b/Sources/ParrotFlow/VocabularyJudge.swift
@@ -41,14 +41,41 @@ enum VocabularyJudge {
         /// words. Past this many the stage keeps what the decoder wrote and
         /// says so — a menu nobody can read decides nothing.
         var slots = 4
-        /// The menu, not the slot count, is what the model is bad at. Three
-        /// binary slots is eight readings and was the most that measured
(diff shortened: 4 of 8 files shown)


znat:
@greptileai review

znat:
Round 1, both findings addressed — one fixed, one partly fixed and partly
argued.

**Issue 1 — persistent transcript-content logging. Fixed** (`f7e7d1e`).

You are right that the line was new disclosure and not only more of what was
already there. The other vocabulary log lines fire when the pass *did*
something; this one fired on any run that produced a slot, including the runs
where the menu is declined and nothing is offered at all.

The count stays unconditional, because a count is the thing `max_slots` is
compared against and a number is not a name. The words moved behind
`PARROTFLOW_JUDGE_DUMP`, which already means "write this dictation's menu down
for a harness to read", so every measurement in the PR body still produces them
and nobody else's log does:

```
$ PARROTFLOW_JUDGE_DUMP=… ParrotFlow --transcribe 17-39-40.wav
vocabulary judge: 5 slot(s) from 6 proposal(s) — "Matthieu" (Matthieu), "went to the" (Matthieu), "Vercel" (Vercel), "deployed on" (Claude), "Vercel" (Vercel)

$ ParrotFlow --transcribe 17-39-40.wav
vocabulary judge: 5 slot(s) from 6 proposal(s)
```

The cap's own line — `"universal" is a Vercel reading past max_per_term 2; not
offered` — is deliberately left ungated. It reports something the app *did* to
the person's transcript, which is the same class as
`"watching" -> "Matthieu" dropped, audio prefers what was written by 3.10`
above it. A withheld reading with no reason in the log is the failure mode this
file spends most of its comments avoiding.

**Issue 2 — no deterministic test for the cap's survivor ranking. Half fixed,
half not reachable.**

Fixed half: a refused fixture, `tests/pipelines/refused/vocabulary-cap-zero.yaml`,
so `check-pipeline.sh` (57/57, in CI) now covers the validation contract — a
cap below 1 silences the stage on every transcript and is refused by name
rather than run.

The ranking itself has no deterministic surface in this repository, and it is
worth being precise about why rather than promising a test that cannot exist
here:

- There is no Swift test target. `Package.swift` declares none, and every check
  in CI drives the built binary.
- The two surfaces that run the judge without audio, `--pipeline` and
  `--replace`, cannot reach the ranking. With no acoustic pass every part is
  `.rule`, so `Standing` never has two values to order. And they cannot even
  produce two rule slots for one term: `ruleParts` offers a rule's substitution
  back only when the term stands **exactly once** (PR #63, round 3), so a
  second occurrence refuses the first as well.
- The input that does exercise all four sources is a wav of Nathan's voice.
  `scripts/check-no-voice.sh` is in CI precisely to keep those out of the
  repository.

So the ranking is measured rather than asserted: `menu-recall.py --runs 3` and
`before-after.py --runs 3` over 37 labelled clips, with the per-clip verdicts
diffed against a baseline measured on `main` the same day — 0 clips moved in
either harness — plus the `max_per_term` sweep in the PR body, which is what
shows the ranking is doing something rather than nothing (1 costs three cases,
2 and 3 and off are identical). Two of those clips exercise the multi-source
case directly: `17-39-40`, where a term holds two rule slots and a spotter slot
and the spotter slot is the one dropped, and `17-39-03`, where a term holds two
scored slots and neither is.

If a Swift test target is wanted, it is a change to `Package.swift` and to CI
that is bigger than this PR and would want its own — happy to open it.

znat:
@greptileai review

znat:
@greptileai review

znat:
Round 2: Greptile re-reviewed `f7e7d1e` — 8 files reviewed, 0 comments added. Both round-1 findings are addressed above; no further actionable findings. Not merging — waiting on you.

PR #67, opened 2026-08-08
test: 90 unfiltered clips, and F15 in the plan

znat:
Two jobs, both text only. No code, no script, no Swift file changed.

## Job 1 — the hand-labelled random sample

`mine-menu-cases.py` selects a clip only when the app's own text already
holds a term, a known mis-rendering, or a word 0.55 similar to a term. A name
the app missed completely cannot enter that way, so the case set is
structurally blind to the misses that matter most (F11). PR #65 showed it:
the `Versailles` → `Vercel` acoustic reach moved from -5.28 to -2.28 and not
one case-set clip changed.

`mine-menu-cases.py --random` sampled 93 clips out of 1579 with no filter at
all. Nathan corrected the labels by hand against the audio. This merges them.

### Counts

| | |
|---|---|
| entries in the sheet | 93 |
| `said:` blank | 0 |
| hand-corrected (`said` differs from the app's own text) | 22 |
| mention a vocabulary term, exactly | 17 |
| already in `tests/menu-cases.yaml`, skipped | 2 |
| dropped | 1 |
| **merged** | **90** |

The sheet did not use the `# picked up:` comment to carry the app's text. It
used a `was:` key instead, one per entry, all 93 present. `was:` is the app's
own output and `said:` is the hand label. The merge turns `was:` into a
`# app:` comment, and writes it only where it disagrees with `said:`. That
matches the file's comment style and keeps it out of the loader's way.

What the 90 merged clips are:

| Class | Count |
|---|---|
| the app missed a term | 3 |
| the app wrote a term over an ordinary word | 2 |
| the app got a term right | 12 |
| no vocabulary term at all, a control | 73 |

The trigger filter would have admitted 23 of the 90. The other 67 it never
sees. One of those 67 is about a term and no filter or floor could ever reach
it: `2026-08-05T23-31-30`, where `Vercel` came out as **"Brazil"**. That clip
is the reason this sample exists.

Two of the merged clips reproduce F5's sentences from the archive:

- `2026-08-08T01-02-23` — app "You don't need to **Supabase** the design",
  said "you don't need to update the design"
- `2026-08-08T00-14-39` — app "I mean in **Redcrawl** in our data", said
  "I mean in general in our data"

### What was dropped, and why

**1 dropped.** `parrotflow-2026-08-06T16-47-02.wav`. The label reads "Okay,
just go back to PR 1 PR one and install." It renders the same spoken words
two ways in one sentence, so it cannot be ground truth for an exact match.

**2 skipped as duplicates.** `parrotflow-2026-08-07T15-36-40.wav` and
`parrotflow-2026-08-07T16-09-50.wav` were already in the file. Their labels
in the random sheet are identical to the ones already there, so nothing was
lost. Dedupe is by wav filename.

**Nothing else was dropped**, and that is deliberate. The 73 clips with no
vocabulary term are the control that shows the pass leaving ordinary speech
alone. Some of them carry decoder errors the vocabulary cannot fix at all
("mermid" for "mermaid", "Z" for "Zod", "grapile" for "Greptile"). Dropping
those would keep only the clips the app can get right or the vocabulary can
fix, which is exactly the selection bias F11 names.

### Verification

No measurement was run. Another agent held the app and Ollama, so this PR
does not build, run, or replay anything.

Mechanical check only: `load_cases()` imported from `scripts/menu-recall.py`
with `SourceFileLoader` and called on the merged file.

```
total cases   : 130
labelled      : 127
unlabelled    : 3
unique wavs   : 130
duplicate wavs: []
comment leaks : []
clips missing : 0
```

Every entry parses. No wav appears twice. Every clip named exists in
`~/Recordings/ParrotFlow Dev`. The 3 unlabelled clips are the pre-existing
blank ones in block 1.

**The recall baseline will move.** `recall 30/37` and `picked 27/37` were
measured on the 37-clip set. The set is now 127 labelled clips. Both numbers
should be expected to fall in absolute terms, because 73 of the new clips are
ordinary speech and some of them the vocabulary can never fix. That is the
set growing, not a regression. Whoever measures next re-records the baseline.
The doc is updated to say so and the PR 3 gate now points at "must not fall
against a baseline measured on the same set" instead of at fixed numbers.

## Job 2 — F15 in the plan

Added to the findings ledger:

> **F15 — a correctly-decoded term gets no slot, so a dropped possessive is
> unrecoverable.** Measured live on 2026-08-08 at 14:37:21 and 14:37:41.
> "Let's praise Matthieu's work" was decoded as "Let's praise Matthieu work",
> and the run produced **no vocabulary log lines at all**. The decoder
> spelled the term right, so nothing was proposed. With no proposal there is
> no slot, the judge never runs, and the `'s` the decoder dropped cannot be
> put back. The possessive variants added in PR 6 only exist where a
> substitution is proposed. Contrast 14:38:21: "Let's praise Praisy's work"
> did produce a proposal, and the judge chose the possessive reading
> correctly.
>
> *This is not judge quality.* At 14:39:15 the judge was offered "Matthew at"
> → `Matthieu` (span 0.52) and → `Matthieu's` (span 0.56), and kept "Matthew
> at". That is the judge picking wrongly, and it is PR 7's territory. F15 is
> the case where the judge is never asked. → PR 9.

F15 goes to a new **PR 9 — a term the decoder got right, missing its
possessive**, added to the build order table and written as a section after
PR 8. Two options are sketched and neither is chosen: (a) open a possessive
slot when a term already in the transcript is followed by a bare noun and the
audio shows an `'s`, or (b) treat it as dictation grammar outside the
vocabulary pass. The section says plainly that PR 9 is unscoped and
unmeasured, and that it blocks nothing.

Two standing regression sentences added:

7. "Let's praise Matthieu's work" — the `'s` must survive (F15)
8. "Let's praise Antonio's work" — the control. Antonio is not a vocabulary
   term, so it shows whether a fix for 7 over-fires on ordinary names.

### DIFF
--- a/docs/proposals/vocabulary-v2.md
+++ b/docs/proposals/vocabulary-v2.md
@@ -76,7 +76,7 @@ Rules for the agent doing a PR. Follow them exactly.
 | Installed app | `/Applications/ParrotFlowDev.app/Contents/MacOS/ParrotFlow` |
 | Judge model | Ollama, `gemma4:e4b`, temperature 0. Do not benchmark with other models loaded; e4b is the stable reference. |
 | Menu cache | `tests/judge-menus.json` (33 menus, 28 reachable) |
-| Case set | `tests/menu-cases.yaml` (37 labelled clips) |
+| Case set | `tests/menu-cases.yaml` (127 labelled clips, in two blocks) |
 
 ### Environment preflight (run before any measurement)
 
@@ -104,8 +104,8 @@ is the prototype; `(CTC-vs-CTC: …)` alone is v1.
 
 | Measurement | Command | Baseline |
 |---|---|---|
-| Menu recall | `python3 scripts/menu-recall.py` | recall 30/37 |
-| Judge took it | same run | picked 27/37 |
+| Menu recall | `python3 scripts/menu-recall.py` | recall 30/37, on the old 37-clip set — stale, see below |
+| Judge took it | same run | picked 27/37, same set — stale, see below |
 | Before/after | `python3 scripts/before-after.py` | 16 fixed / 10 broken / 2 regressed |
 | Judge on cache | `python3 scripts/tune-judge.py` | 24/28 with the sentinel bug; 26/28 with sentinel lines stripped |
 | Two-stage (fixed harness) | see PR 2 | 25/28, ceiling 28/28 |
@@ -114,6 +114,14 @@ Gates for every PR from PR 3 on: recall must not fall; `before-after.py`
 must show no new REGRESSED rows; judge `picked` must not fall below the
 baseline recorded for the same cache. A `--harvest` re-run changes the cache;
 re-record the baseline in the PR when you re-harvest, and say so.
+
+**The recall and picked baselines are stale.** They were recorded on a
+37-clip set. `tests/menu-cases.yaml` now holds 127 labelled clips: 90 more
+were merged from an unfiltered random sample (F11). The next run re-records
+both numbers. Expect them to drop in absolute terms, because 73 of the new
+clips are ordinary speech with no vocabulary term in them, and some of those
+carry decoder errors the vocabulary cannot fix. The gate is still "must not
+fall" against a baseline measured on the same set.
 
 **Caveat that stands until PR 1:** the same clip scores ~12 nats apart
 between live dictation and replay. Every number above describes replays.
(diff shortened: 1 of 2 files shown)


znat:
Both findings fixed in 9675818.

**P1 — aggregate gate masks vocabulary regressions.** Agreed, and it is worse than it looks: the set went from 37 clips that were nearly all about a term to 127 clips of which 73 are controls. The gate is now per class. No clip that is about a vocabulary term may get a worse mark, and the 73 controls are reported as a separate total that should not move in either direction. `menu-recall.py` prints one pair of totals and does not know which block a clip came from, and this PR does not touch scripts, so the doc says splitting the report belongs to whoever next touches PR 2's harness. Until then the split is by hand: the `# picked up:` comment on every block 2 entry names the class, and every block 1 clip is about a term.

**P2 — F15 overstates its audio ground truth.** Correct. What was measured is the transcript and the empty log. Nobody listened to the clip. F15 now says that, and separates the two claims: the sentence was meant as a possessive because Nathan dictated it, but whether the `'s` is audible is unknown. PR 9 now starts by checking the audio, because option (a) only works if the sound is there.

@greptileai review

PR #69, opened 2026-08-08
test: four ways to ask the judge, measured

znat:
A measurement spike. **No prompt, no app and no schema changed.**
`examples/prompts/verify_names.md` and `vocabulary.yaml` are untouched.

The judge scores 41/53 on the cached menus. Its errors concentrate in one
class: an ordinary English word collides with a vocabulary term and the term
wins. PR #68 recorded nine such clips. **On those, the shipped prompt scores
0 of 8** (the ninth, `09-35-01` "the Versailles castle", never held the true
sentence on its menu, so it cannot be scored).

This PR asks whether the *question* is wrong, not its wording.

**Read [`docs/proposals/judge-framings.md`](docs/proposals/judge-framings.md)
first.** It is the short version: the verdict, the actual text of every arm as
a diff on the shipped prompt, and three real menus from the cache. The tables
below are the full record.

## How it was measured

`scripts/judge-framings.py`, offline against `tests/judge-menus.json` — 66
menus, 53 reachable, chance **17.4/53**. Judge `gemma4:e4b` at temperature 0,
Ollama; `gemma4:e4b-mlx` and `gemma4:12b` unloaded with `keep_alive: 0`, so
only `gemma4:e4b` was resident (F8, and "isolate models when benchmarking").
No build, no install, `~/.config/parrotflow-dev/` untouched.

Every framing is an exact edit list applied to the shipped prompt, and `build`
raises if an edit stops matching it exactly once. A framing cannot quietly
become the baseline.

**I tuned and reported on the same 53 cases. There is no held-out set.**
F16 measured five wordings of one sentence spanning 38–42, so anything inside
±2 has not moved.

**Repeatable.** `baseline`, `typed-slot` and `no-terms-cut` each ran twice in
separate processes and reproduced exactly — same total, same per-case diff.
Baseline reproduced 41/53, as PR #68 recorded.

## Every framing, against chance 17.4/53

| # | framing | wordings | picked /53 | collision class /8 |
|---|---|---|---|---|
| 1 | baseline, the shipped prompt | — | **41** | **0** |
| 2 | no term list, enumeration neutralised | 1 | 36 | 3 |
| 2 | no term list, enumeration deleted | 1 | 38 | 4 |
| 3 | inverted polarity, ported from `misheard` | 1 | 35 | 1 |
| 3 | inverted polarity, short | 1 | 35 | 0 |
| 3 | inverted polarity, long | 1 | 38 | 1 |
[...]

PR #71, opened 2026-08-08
test: route the case away from the judge, and measure what it costs

znat:
**The router loses. Do not build it.** Control 41/53, best routed arm 32/53,
chance 17.4/53. Measurement only — no app change, no prompt shipped.

Round 1 (#69, F17) found the judge fails one class because the term list has to
be present for a name to be spellable and absent for a name to be doubted. This
tests the repair: send each case to one decider instead.

- decoded word is **not** a real word → decide in code, better raw score wins,
  no model call
- decoded word **is** a real word → ask a judge one question about position,
  no term list

`Replacements.isRealWord` is the router under test. It is reproduced offline
by `scripts/real-words.swift` (the same NSSpellChecker call, `en` then `fr`)
and cached in `tests/real-words.yaml`. **This reproduces the app's gate; it
does not call it** — a Python harness cannot link the app bundle.

## The split

77 uncertain slots over the 53 reachable cached menus.

| slots | class | argmax raw score |
|---:|---|---|
| 8 | not a word, and scored | **0/8** |
| 5 | a dictionary word, capitalised mid-sentence | 5/5 |
| 27 | a dictionary word, lower case | 15/27 |
| 17 | a span of more than one word | 8/17 |
| 20 | no score line at all | cannot run |

**8 cases carry no score block at all.** 16 hold at least one slot with no
score line. Those cannot be routed by score under any rule.

By case: **2 go to code, 51 go to the judge.** A case goes to code only when
every one of its slots can.

## Branch B — 0/8, and it is not bad luck

`Vocabulary.autoApplies` writes a term in when the term beats the decoded word
on the raw score **and** the decoded word is not a real word. So a slot reaches
a menu only when one of those failed. On the not-a-word class the second test
passed, so the first must have failed — the decoded spelling already scores at
least as well as the term. It does, on 8 of 8. Argmax there can only answer
"keep what the decoder wrote", and it is 0/8 because in all 8 the speaker meant
the term. The judge sees the residue of a decision code already took.

A sweep of a constant handed back to the term (`--sweep`, no model calls)
reaches 8/8 on that class at about 1.5 nats. It proves nothing: the class is
unanimous, there is no clip where a non-word decoding was right, so no rule
there can fail. Meanwhile on the dictionary-word class the constant policy
"keep what was decoded" scores 22/32 and argmax scores 20/32 — the class label
alone beats the number.

## Branch A — 51 cases, chance 16.4

Two wordings of the position question, each as a whole-sentence menu and as one
blank per slot. Four judge configurations plus the shipped control.

| arm | branch A /51 | multi-slot /18 | combined /53 | collision /8 |
|---|---|---|---|---|
| shipped (control) | **39** | 12 | **39** | 0 |
| position, sentence | 24 | 4 | 24 | 3 |
| position wording 2, sentence | 32 | 6 | 32 | 4 |
| position, blank per slot | 31 | 7 | 31 | **6** |
| position wording 2, blank | 27 | 6 | 27 | **6** |

No router at all: **41/53**.

Wording moves 8 cases in the sentence form — more than the form moves anything
— so the totals turn on wording, not on form. But on the collision class the
same wording goes 3/8 as sentences and 6/8 as blanks, and the three it gains
are exactly the multi-slot clips a whole-sentence menu cannot express:
`17-47-45`, `14-04-21`, `10-12-37`. The last is the clip F17 named as the
argument for a per-slot question.

## The routing errors — 12 of 53

**Sent to code, and code got it wrong (2 — the whole of branch B):**

```
17-26-56  I changed settings in Vercel          Versal. vs Vercel.  sound prefers Versal. by 1.5
09-46-57  The app is being deployed on Vercel.  Versal. vs Vercel.  sound prefers Versal. by 0.9
```

**Sent to the judge on a word nobody dictating could have meant (10):**

```
16-09-50  Let's praise the work Praisy has done.        Prissy  vs Praisy
16-09-37  Let's praise the work Praisy has done.        Prissy  vs Praisy
16-09-15  Let's praise the work Praisy has done         Priss.y vs Praisy
16-09-00  Is Praisy here today?                         Prissy  vs Praisy
15-36-50  Let's praise the work that Praisy has done.   Prissy  vs Praisy
15-36-40  Praisy is here.                               Prissy  vs Praisy
15-21-19  Praisy did a great job.                       Prissy  vs Praisy
13-19-22  Praisy is with us today.                      Prissy  vs Praisy
14-11-48  Praisy sent me the Ollama benchmarks ...      Prissy  vs Praisy
15-21-29  Matthieu published a pull request.            Mathieu vs Matthieu
```

`Prissy` is in the English dictionary. `Mathieu` is in the French one, and the
app asks both. Argmax would have been right on all ten. The gate is close to
inverted: the two cases it sends to code are the two it should not, and ten of
the cases it protects from code are the ones code would have settled.

A second router — a capital mid-sentence overrides the dictionary — was tried
because that list is long. It moves branch B from 2 cases to 3 and misroutings
from 12 to 11. Only 5 of the 27 offending slots carry a mid-sentence capital,
so it fires on a fifth of its own class.

## What I recommend

1. **Do not build the router.** The design assumes the acoustic score is
   informative where language is not. It is the reverse here.
2. **Measure the blank form next, carrying the shipped question.** This spike
   changed form and content at once, so it cannot say what blanks are worth on
   their own. The collision-class result says they are worth measuring.
3. **Fill the gap in `menu-cases.yaml`.** No clip exists where a non-word
   decoding was what the speaker said. Until one does, no rule on that class
   can be falsified.

## Method

`scripts/judge-routing.py` against `tests/judge-menus.json`, `gemma4:e4b`,
temperature 0, nothing else loaded. The shipped control was re-measured in the
same run at 41/53, reproducing round 1 exactly. Three full runs gave identical
totals and identical per-case diffs. No reply lacked a bare letter. No build,
no install, no `~/.config/parrotflow-dev/`.

**The same 53 cases are tuned on and reported on. There is no held-out set.**
The second router was chosen after seeing the first one's errors.

Slots are recovered by diffing the menu options against each other, because
`VocabularyJudge` throws the combination away. The recovery is checked: the
slot options must rebuild the menu exactly, and they do for all 53.

`scripts/tune-judge.py` gains one optional `lead` argument so the blank form
can put the sentence above the lettered options and still have its reply read
by the app's own rule.

Report: `docs/proposals/judge-framings.md` (round 2 section), finding F18 in
`docs/proposals/vocabulary-v2.md`.

### DIFF
--- a/docs/proposals/judge-framings.md
+++ b/docs/proposals/judge-framings.md
@@ -269,3 +269,313 @@ sentences cannot express. It also makes the cost linear in slots instead of
 the product PR 6 had to cap at two.
 
 Recorded as F17 in [vocabulary-v2.md](vocabulary-v2.md).
+
+---
+
+# Round 2 — routing the case away from the judge
+
+**The router loses. Do not build it.**
+
+Round 1 said the judge fails one class because the term list must be present
+for a name to be spellable and absent for a name to be doubted. Round 2 tested
+the obvious repair: send each case to only one of them. If the decoded word is
+not a real word, decide in code on the raw acoustic score. If it is a real
+word, ask a judge one question about position, with no term list.
+
+Both halves fail, and they fail for different reasons.
+
+- **The code half cannot win, by construction.** `Vocabulary.autoApplies`
+  already applies argmax on the raw score. A slot only reaches a menu when
+  argmax said no. On all 8 not-a-word slots in the cache the decoded spelling
+  scores at least as well as the term, and on all 8 the speaker meant the
+  term. **Argmax scores 0/8.**
+- **The judge half loses 7 to 15 cases.** The best position arm scores 32/51
+  against the shipped prompt's 39/51 on the same subset.
+- **Combined, the best arm is 32/53 against 41/53 today**, chance 17.4.
+
+One thing did move, and it is about form, not about the question. Asking per
+blank instead of per sentence takes the collision class from 3/8 to 6/8 at the
+same wording, and wins exactly the three multi-slot clips a whole-sentence
+menu cannot express. It loses too much elsewhere to ship.
+
+Measured with `scripts/judge-routing.py` against `tests/judge-menus.json`,
+`gemma4:e4b`, temperature 0, nothing else loaded. The shipped control was
+re-measured in the same run at **41/53**, reproducing round 1 exactly. Three
+full runs gave identical totals and identical per-case diffs. No reply lacked
+a bare letter. No build, no install.
+
+**The same 53 cases are tuned on and reported on.** There is no held-out set.
+The second router below was chosen after looking at the first one's errors.
+
+## What was tried
+
+**Two wordings** of the position question, each in **two forms** — whole
+sentence and one blank per slot. Four judge configurations, plus the shipped
+prompt as control. **Two routers**: `Replacements.isRealWord` alone, and the
+same with a capital mid-sentence overriding it.
+
+## Recovering the slots
+
+The cache stores whole sentences. `VocabularyJudge` composes them from a
+cartesian product over uncertain spans and throws the combination away, so
+`judge-routing.py` recovers the spans by diffing the options against each
+other. The recovery is checked: the slot options must rebuild the menu
+exactly, and they do for all 53. Five menus hold two adjacent spans that no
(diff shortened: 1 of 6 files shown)


znat:
@greptileai review

PR #82, opened 2026-08-09
test: how much of the veto one bad clip destroys (PR 6a)

znat:
## Summary
PR 6a of the plan asks whether one mislabeled clip in the reference bank can quietly wreck the veto's accuracy, and the plan's own falsifier for this was written against a statistic (AUC) that can't see the effect at all.
This PR measures the veto's actual true/false rejection counts under a poisoned clip, across eleven terms, instead of only AUC.
One bad clip drops true rejections from 554 to 228 while buying back only 8 of 10 false ones — 40 true rejections lost per false one saved — so the falsifier's condition fired and its AUC-based conclusion was wrong.

## Issue
Part of #74. Stacks on #80 (`docs/missing-prs`), the tip of the stack.

## What it measured

**The falsifier's condition fired, and its conclusion is false.** Nine of eleven terms show no AUC movement at all under poisoning — that's arithmetic, not evidence, because a term's spread is a constant that divides out of any ranking inside that term. The veto's actual decision can see it, and it collapses: true rejections go from 554 to 228 across the eleven terms, while only 8 of 10 false rejections are bought back. `Arexvy` goes from catching 83% of the spans it should catch to 2%.

## What is new

**The two clips the plan named as the likely culprits are not the clips that actually set the spread.** Both terms' actual maximum-distance clips are real recordings of the term, not mislabeled ones. Removing one of the named clips *raises* its term's spread, because it was another recording's nearest neighbor. A bad clip with company never sets a maximum, so a pruner that just deletes the farthest recording would delete a real one.

**The 90th percentile is worth building instead of the maximum.** It keeps 551 of 649 true rejections under the same poison (against the maximum's 228 of 554), at a cost of seven more false rejections out of 96. It stops being robust below about eight recordings per term.

**The second-pronunciation arm could not be run, and was not faked.** No observation in the archive carries a language tag distinguishing pronunciations, so this part of the plan (§7) stays unmeasured — it needs a dedicated recording session.

## How

`scripts/reference-matching.py` gets a `--poison` arm. Deterministic: no app, no build, no decoder, no model. Every distance is computed once and cached, so each arm after the first is just an index lookup. The baseline reproduces round 7 exactly (`Supabase` 1.000, `Redcrawl` 0.966, pooled 0.935).

The measurement is in-sample, and the PR says so: recordings were mined from the clips they're scored on. The per-clip hold-out is applied throughout but doesn't cover the corpus-level fit — the deltas between arms are the point, not the absolute numbers.

<details>

No Swift changed. Nothing under the live config or voice store was written — the arms run on a copy under `PARROTFLOW_CONFIG_DIR`. No wav and no transcript is committed; `scripts/check-no-voice.sh` passes 5/5.

</details>

### DIFF
--- a/docs/proposals/vocabulary-v3.md
+++ b/docs/proposals/vocabulary-v3.md
@@ -544,6 +544,17 @@ term. **One bad clip disarms the veto for that whole term.** The code's own
 comment says so and names two, both mined automatically:
 `Vercel/09-brazil.wav` and `Tasmeen/06-that'smeanssend.wav`.
 
+**Measured, 6a.** The mechanism is real and it is large. One bad clip injected
+into each of the nine clean folders takes the veto from 554 true rejections to
+228 across the eleven terms, and widens the tightest banks by about half a unit
+— `Supabase` 2.919 to 3.430, `Arexvy` 3.031 to 3.603. **But the two clips named
+above are not the clips that set the spread.** `Tasmeen`'s maximum is held by
+`01-tasmin.wav` and `Vercel`'s by `13-versal.wav`, both real recordings of the
+term. `06-that'smeanssend.wav` is 5th of 8 and `09-brazil.wav` is 8th of 16.
+Removing them changes nothing and makes it worse: taking `09-brazil.wav` out
+*raises* `Vercel`'s spread, 3.178 to 3.235. A bad clip with company never sets a
+maximum. The result block under 6a has the numbers.
+
 **The query side is far safer, and that asymmetry matters.** `distance` is a
 `min`, so a bad clip only counts when it is the nearest thing to the span being
 judged. A bad clip is by definition unlike the term, so for a genuine utterance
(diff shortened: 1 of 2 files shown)


znat:
Round 1 addressed in fa7838c. Both comments were valid.

**Cache accepts mismatched matrices.** Fixed. The key was the recording names and the span count. It is a sha256 over the MFCCs themselves now, each with its row's identity — name, term, source clip for a recording; clip, set, term, group for a span. Verified three ways: a recording perturbed by 1e-9 rebuilds, two spans swapped rebuild, unchanged input hits.

**Missing injection crashes poison arm.** Fixed. `--inject` is checked against the exemplars in scope before any matrix is built, and it names both flags:

```
✗ no recording Vercel/09-brazil.wav under source scripted.
  63 recording(s) are in scope. `--source all` keeps every one of them.
```

One more of the same shape, not reported: injecting a recording the folder already holds put the same index in the bank twice, which makes a leave-one-out distance meaningless. That arm is now skipped and says so.

No measurement changed. Both arms reproduce byte for byte against the run in the result block.

znat:
Round 2 addressed in 6c5940c.

**Legacy caches crash validation.** Valid, fixed. The check now requires `stamp`, `se` and `ee` all to be present before it trusts any of them, so a cache written in the round 1 format rebuilds rather than raising. Tested against exactly that: a npz with `names`, `spans`, `se`, `ee` and no `stamp` prints `the cache does not match this audio; rebuilding` and carries on.

**The other two comments were re-anchored from round 1 and both were already fixed in fa7838c.** The cache is keyed on a sha256 of the MFCCs with each row's identity, so reordered spans and replaced audio at the same path both rebuild. `--inject` is checked before any matrix is built. To take the flagged shape out of the code entirely, the lookup is now hoisted to the line after its guard:

```python
poison = index[injected]           # checked above; never looked up again
```

No bare `index[injected]` is left in the loop.

No measurement changed in either round. Both arms still reproduce byte for byte against the run in the result block.

PR #83, opened 2026-08-09
docs: the architecture the evidence points at, and five PRs for it

znat:
## Summary
The plan's clip-bank findings, scattered across several PRs, implied an architecture nobody had written down, and two open questions — does `acoustic: false` also disable the veto, does the clip bank ever gate a `heard:` rule — were unverified.
This PR documents four facts checked against the code, then proposes five new PRs (11–15) for a per-term, evidence-gated architecture.
Every new item is explicitly marked as a proposal, not a result — nothing here is measured, and PR 1 and PR 6 are flagged as mutually exclusive as currently written (one needs the acoustic path off, the other needs it on), a gap the plan never stated.

## Issue
Part of #74. Stacks on #82 (`test/poison-a-clip`).

## Four facts, verified against the tree

1. **The clip bank is wired inside the acoustic pass**, so `acoustic: false` switches the veto off too. This means PR 1 and PR 6 are mutually exclusive as both are currently written, which the plan never said.
2. **The clip bank has never gated a `heard:` rule substitution, in any arm.** Rules apply in the pipeline before the acoustic pass runs. Every clip-bank number in the plan so far is a number about acoustic proposals, not about rules.
3. **`heard:` matching is exact, whole-word, and case-insensitive.** Every inexact route lives in the acoustic pass alone. This is the mechanism behind #79's `Praisy` loss.
4. **The judge is mostly not discriminating.** Multiple examples show it picking the same answer at every slot regardless of which is right, which is what "always keep the term" would also produce.

## The architecture, and what changed under it

A new direction (not a result): the spotter generates and decides nothing; the clip bank decides, from positives and negatives; the `heard:` map covers a term's first days; the judge is the fallback — each switched on per term by evidence.

**PR 6c (clip review) is rewritten, not replaced**: the outlier-hunt approach is falsified (bad clips arrive in groups and vouch for each other), replaced with grouped review that never deletes, only marks a group not-counted.

**PR 6d (bank as proposer) is reframed**: since the spotter is a good detector (0.999) but a bad chooser (0.318), take the choosing away from it and measure the false-positive rate first, with a no-generator arm as blind control.

**PR 6e (abstain curve) is new** — subsample each term's bank and plot per-term AUC against clip count, correcting the record on the earlier "about three corrections" claim, which came from one term at one count and shouldn't be quoted as general.

## Five new PRs

| | |
|---|---|
| **PR 11** | negative clips as a data type, with explicit polarity and `collides_with:` |
| **PR 12** | a revert never deletes a sample — blames the logged proposal, reduces `seen:`, counts the revert |
| **PR 13** | the clip bank as a gate on `heard:` rules — veto only, not confirm |
| **PR 14** | the `heard:` map is a bootstrap with no cutover |
| **PR 15** | the judge on probation — the blind control nobody has run: `acoustic: false` with and without the judge stage |

Each carries Changes / Why / Size / Verified by / Falsified if, and each is explicitly marked unmeasured.

<details>

Documentation only, no Swift. Three line references in the working notes were off by a few lines and are corrected in the text; the substance held in all three. `scripts/check-no-voice.sh` passes 5/5.

</details>

### DIFF
--- a/docs/proposals/vocabulary-v3.md
+++ b/docs/proposals/vocabulary-v3.md
@@ -2344,9 +2344,48 @@ is the version with no generator in front of it, and it is the blind control
 part 1 §2 asks for: if the bank behind the spotter does no better than the bank
 on its own, the spotter is contributing nothing.
 
-**Changes.** Offline first, in the same script 6a extends
-(`scripts/reference-matching.py` on `origin/spike/exemplars-round-2`). No app
-change until the false-positive number says there is a point.
+**Where the raw hits come from, and how they can be got wrong.** This has to be
+stated, because there are three different populations here that all look like
+"the spotter's output" and measuring the wrong one is silent.
+
+The population 6d wants is `spotted.detections` as
+`spotKeywordsWithLogProbs(audioSamples:customVocabulary:minScore: nil)` returns
+it (`Vocabulary.swift:551-553`) — every hit, unfiltered. What must **not** be
+substituted for it:
+
+- **Post-gate spans.** `acousticSpans` applies three gates before anything is
+  proposed — the best term per stretch of audio, nothing already proposed, and
+  nothing sitting on an ordinary word (`Vocabulary.swift:788-792`). Those gates
+  are part of what 6d is replacing, so their output is the wrong population.
+  `Vocabulary.spotterFloor` is a fourth filter on the same output.
+- **`--spot`'s detections.** `SpotCommand` calls the same method with
+  `minScore: -20` and builds its context from `--terms` or the bare term list,
+  where `Vocabulary.prepare` registers the pronunciations too. Same shape,
+  different population.
+- **Word spans from `trace.jsonl`.** Those are the second arm, the blind
+  control. Not the first.
+
+**What exists today and what does not.** `PARROTFLOW_SPOTTER_DUMP=1` writes
+every detection as `spotter: <term> <score> at <start>s-<end>s`
+(`Vocabulary.swift:559-566`), and the decoder's own word spans as
+`word <text> <start>-<end>` (`Vocabulary.swift:540-544`). That is the right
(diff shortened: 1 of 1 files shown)


znat:
**Greptile round 1 — addressed in 7f7d415.**

> **Raw spotter input is unspecified.** The sweep requires every pre-gate spotter hit, but assigns the work to an offline Python script and only identifies `trace.jsonl` as the source for the separate word-span control.

Valid, and the failure would have been silent. 6d now names the population and
the three things that must not stand in for it.

**The population it wants:** `spotted.detections` as
`spotKeywordsWithLogProbs(..., minScore: nil)` returns it
(`Vocabulary.swift:551-553`) — unfiltered.

**What must not be substituted:**

- **Post-gate spans.** `acousticSpans` applies three gates before anything is
  proposed — best term per stretch, nothing already proposed, nothing on an
  ordinary word (`Vocabulary.swift:788-792`) — and `spotterFloor` is a fourth
  filter on the same output. Those gates are part of what 6d replaces.
- **`--spot`'s detections.** Same method, but `minScore: -20` and a context
  built from `--terms` or the bare term list, where `Vocabulary.prepare`
  registers the pronunciations too.
- **`trace.jsonl` word spans.** That is the second arm, the blind control.

**What exists and what is owed.** `PARROTFLOW_SPOTTER_DUMP=1` already writes the
right population (`Vocabulary.swift:540-544`, `:559-566`), but to the app log,
which truncates at 1 MB — so 141 clips cannot be collected from it after the
fact. 6d therefore owes a dump to a file first, on the
`PARROTFLOW_REFERENCE_DUMP` model (`ReferenceMatch.swift:50-52`, `:276-286`).

6d's "no app change" line is corrected rather than left standing: the dump **is**
an app change. It writes a file and decides nothing.

PR #85, opened 2026-08-09
fix: a revert does not write a rule, and keeps the audio

znat:
## Summary
On `main`, taking a term back (for example `--learn Praisy praise`) wrote a `config.yaml` replacement rule in the wrong direction, silently disabling that term for every future correct use, and `--forget` couldn't clean it up because nothing pointed back at the rule.
This PR refuses that write in `ConfigWriter`, and instead counts down the rule that actually fired, records the collision, and keeps the correction's audio as a negative sample in a separate directory.
`scripts/check-reverts.sh` (67 new cases) passes in CI, no sample is ever deleted by a revert, and a term is never silently disabled by taking it back; a live revert itself is not measured here — it needs a person at a microphone.

## Issue
Part of #74. Targets #81 (`feat/corrections-keep-audio`). Two commits, separable — the bug fix is independent of this stack and cherry-picks cleanly to `main` on its own.

## The bug

`ConfigWriter.addReplacement` wrote `corrected: [heard]` into `config.yaml` regardless of the direction of the correction. So a speaker taking a term back — "the app wrote Praisy, I meant 'praise'" — ended up disabling the term entirely, including every future correct use of it. `--forget Praisy` then removed the pronunciation but left the broken rule sitting in `config.yaml`, unreachable by any command.

## Commit 1 — the fix

`addReplacement` now refuses that shape and returns false. The gate lives in the writer, the last function before the file, so all three call sites are covered and a future call site can't reintroduce the bug.

## Commit 2 — what a revert does instead

1. **Counts down whatever actually fired** — an exact `pronunciations:` rule or the acoustic pass — rather than deleting it. A rendering seen nine times and reverted once is still how this person usually says the word; the entry is only removed at zero, and only if a correction wrote it (never a legacy or mined entry, which has no honest count to take one off).
2. **Records the pair** as `collides_with:` under the term, without ever matching or substituting on it.
3. **Keeps the audio as a negative** in a separate `voice/negatives/<Term>/` directory rather than a flag inside the positive samples directory, so existing scripts that walk the samples folder don't silently ingest a negative as a positive.

No sample is ever deleted by this path — #82 already showed a bad clip can't be reliably picked out of a bank from one revert alone.

**Two counting bugs found while writing the tests, both fixed here:** the sample cap matched files by bare filename, so a negative could have shadowed a positive nobody confirmed and gotten the confirmed file deleted instead; and pruning kept a reverted word's rendering alive past the seen-once rule because it read the wrong key.

<details>

`scripts/check-reverts.sh`, 67 cases, in CI. Each case builds a scratch config directory and generates synthetic tone-burst audio. Covers: no rule written anywhere, the count-down and drop-at-zero behavior, the legacy-entry exception, the negative clip and its polarity tag, `collides_with:` accumulating on repeat reverts, the forward (non-revert) direction still working unchanged, and `--forget` clearing pronunciations, samples, negatives, and collisions together.

CI checks touched: `check-corrections.sh` 59/59, `check-vocabulary-config.sh` 61/61, `check-no-voice.sh` 5/5, `check-replacements.sh` 23/23.

**Not done: a live revert.** It needs a human at a microphone. The panel's own path takes the same route as `--learn <Term> <word> --clip`.

</details>

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -152,6 +152,15 @@ jobs:
       - name: A correction keeps the audio
         run: scripts/check-corrections.sh
 
+      # The other direction, which used to write a rule that disabled the term
+      # it was reverting. A revert must write nothing that can fire on text
+      # later — not to config.yaml, not to vocabulary.yaml — and its clip must
+      # land in negatives/ and never in samples/. Every number this project
+      # computes walks samples/, so a negative filed as a positive would rot
+      # all of them silently.
+      - name: A revert writes no rule
+        run: scripts/check-reverts.sh
+
       # The same trap, one directory over: the transform folder a first launch
       # is seeded with against the copy in examples/ that people read and
       # score. It has always been able to drift and has never been checked
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -1871,13 +1871,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         if let seen = outcome.seen {
             Log.write("    \(term): seen \(seen) time(s), from correction")
         }
+        for line in Corrections.said(about: outcome.revert, term: term) {
+            Log.write("    \(line)")
+        }
         if let sample = outcome.sample {
             Log.write("    kept the audio as voice/\(sample)")
         } else if let why = outcome.skipped {
             Log.write("    no audio kept — \(why)")
         }
+        let bank = outcome.revert == nil ? "samples" : "negatives"
         for gone in outcome.capped {
-            Log.write("    capped voice/samples/\(term)/\(gone.file) — \(gone.why)")
+            Log.write("    capped voice/\(bank)/\(term)/\(gone.file) — \(gone.why)")
         }
         for gone in outcome.pruned {
             Log.write("    dropped \"\(gone)\" from \(term) — seen once and not again since")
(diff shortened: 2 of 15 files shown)


znat:
Both addressed in b3b7b7e.

**P2 — inline normalization discards comments.** Right in substance, though not for the shapes named: `Praisy: [Prissy] # note` and `Mirza: 0.85 # note` both keep their comment today, because the comment travels inside the value. The shape that breaks is the one that *wraps*, which is what the live file writes:

```yaml
  Praisy: [Prissy,   # the common one
           Pressy]
```

Joining those onto one line gives `heard: [Prissy,   # the common one            Pressy]`. Everything after the `#` is comment, the bracket never closes, and `vocabulary.yaml` then loads as **no terms at all** — silently, since a file that will not parse reads as an empty vocabulary and `--check-config` simply prints no term line. Reproduced before the fix.

Now only the first line is rewritten; the continuations move under the new key untouched, still indented deeper than it, which is all a flow sequence asks. This predates the PR — it is on `feat/corrections-keep-audio` in `adding(pronunciation:)` — and it is reachable from both a correction and a revert, so both are scored.

**P1 — revert persistence allows double-decrements.** Right. There is no transaction across `vocabulary.yaml` and `voice/`, so what is available is ordering by what a retry costs. The three writes are now: the observation (a row in an append-only file — a duplicate is visible and harmless), then the collision (a counter — one extra is recoverable), then the count down last (destructive — a `seen:` taken down twice can delete a rendering that should have stood at one, and nothing puts it back). A failure anywhere above now leaves the count untouched and the retry is clean, which removes the case you named exactly.

`check-reverts.sh` is 74/74, up from 67. The seven new cases cover the wrapped-value shape in both directions and check that the file still loads afterwards, which is the assertion that would have caught this.

PR #109, opened 2026-08-11
feat: the offer names commands, and you can click them

znat:
PR 5 of the HUD placement and offer stack. Base is `feat/near-black-surfaces`
(#108), which is not merged — the stack is reviewed unmerged, so this diff is
this PR's work only.

Spec: `docs/proposals/hud-placement-and-offer.md`, section 3. That section
describes the finished offer, which spans PRs 5 to 9. This is the first slice:
a working, clickable offer with **no system-wide event tap in it**.

## What changes

The pill said *"Wrong? Right ⌘ to fix it"*. That was a whole sentence because
the pill used to appear at the bottom of the screen with no connection to the
words it was about. PRs 2 and 3 put it under those words, so the words are the
subject now and the pill only has to name what can be done to them.

One chip per command, each carrying the letter that will run it:

```
[ C Correct ] [ G grammar ]
```

The list is `["Correct"] + transforms where offer: true`. Correct is first and
is not a transform: it is the one command about the words rather than about
rewriting them, it needs no model, and it cannot fail.

## The two new config keys

```yaml
  - name: grammar
    description: fix grammar and punctuation mistakes
    offer: true       # a chip on the pill
    key: g            # the letter drawn on it
```

Both default off. The offer is on screen for a few seconds and every entry
costs the others room, so a transform joins it only by asking. The shipped
`grammar` asks; nothing else in the default config or in
`config.example.yaml` does.

`key:` is cut to one character and upper-cased. `--check-config` now prints
what is on the pill and the letter each chip carries — a chip that failed to
appear, or a `key:` that was cut, is otherwise invisible in three seconds of
pill.

Documented in `docs/pipelines.md` beside `display:`, and the `correct_offer`
paragraph in `docs/configuration.md` is brought up to date.

## Four decisions, not taste

- **A letter, not an icon.** An icon says what a command is *about*; a key says
  you can press it, which is the more useful thing on a surface up for a few
  seconds.
- **Nothing is preselected.** `selected` starts nil. It is only ever the
  pointer's mark and it does not outlive the pointer — leaving the pill clears
  it. Nothing runs without a click, so a chip lit before you have touched
  anything is saying something about a command that is not about to happen.
- **The chips are not `Button`s.** The pill is a `nonactivatingPanel` and is
  never the key window, and SwiftUI draws controls in an inactive window at
  reduced emphasis — a lit chip came up washed out until the pointer landed on
  it, which read as nothing being lit at all. Tap gesture, explicit
  `contentShape`.
- **The pill takes the mouse only in the offer state.** `ignoresMouseEvents` is
  set per state in `set(_:for:)`. The pill sits over whatever you are working
  in, so a surface that swallowed clicks for the length of a dictation would be
  a hole in your screen — but the offer is made of buttons, and a button you
  cannot click is a picture of a button.

## What a click does

**Correct** opens the correction panel over the transcript frozen when the
offer went up — not the preview panel, which is what the base does today. The
offer names one thing and that is the surface that does it.

The sentence goes back into the field frozen with it, through
`replace(_:with:)` — the routine the tapped offer already uses. The panel can
be open for as long as it takes to think about a spelling, and push-to-talk
does not wait for the previous transcript, so anything read at save time can
belong to a newer dictation. Teaching a rule is the panel's own and is
unchanged: `learn(_:)` is split out of `saveCorrections` and shared, and
`saveCorrections` itself behaves exactly as it did. Rebuilding the panel is
PR 11.

**A transform** calls the existing `runTransform(_:instruction:)` with an empty
instruction, so it still shows its preview — that is what `confirm:` asks for
today. Making an offered command rewrite in place without a preview is PR 9.

Correct is identified by its index, not by its title, so a config free to name
a transform "Correct" cannot take that slot over.

## Explicitly not here

| Not in this PR | Owner |
|---|---|
| `OfferKeys`, the event tap, the letters actually working | PR 6 |
| decay, fade, hover holding the clock | PR 7 |
| when the offer is raised | PR 8 |
| an offered command rewriting in place, no preview | PR 9 |
| the correction panel rebuilt around spans | PR 11 |

`takeTheOfferIfTapped` — the hotkey tap — is untouched and still opens the
preview panel. PR 6 is the one PR that swallows keys from every app on the
machine, and it should be reviewed as exactly that and nothing else.

One deliberate deletion: `showCorrectOffer` no longer requires
`hotKeys.binding?.displayName`. That guard existed only because the pill
printed the hotkey's name, and it does not print it any more.

## Checks

- `swift build -c release` clean. 20 compiler warnings before, 20 after — all
  the pre-existing `AppDelegate` Sendable ones, none new.
- `swiftlint` reports nothing new in the changed files.
- Every deterministic script from `.github/workflows/checks.yml` run locally
  and green, including `check-default-config.sh` (28 keys, 9 transforms) and
  `check-vocabulary-config.sh` (61/61). `check-inplace.sh` and `check-span.sh`
  were **not** run: they replace `/Applications/ParrotFlowDev.app` and take over
  the keyboard.
- `--check-config` against both the shipped default and `config.example.yaml`
  reports `offer  1 transform(s) on the pill, after Correct / grammar — G`.

## Please render the sheet

`--panel-sheet` is updated for the new `.offer` shape, and `--panels offer` now
wires hover and click so the state can be worked with a pointer. A human should
look at it:

```sh
make app && .build/ParrotFlowDev.app/Contents/MacOS/ParrotFlow --panel-sheet /tmp/sheet.png
```

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -2081,17 +2081,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         //
         // The correction panel rather than the preview panel — the offer names
         // one thing and that is the surface that does it. Over the sentence
-        // frozen when the offer went up, so a dictation that finished in the
-        // meantime cannot move what is being corrected. Saving is the panel's
-        // own, unchanged: the same `saveCorrections` the spoken command uses.
-        guard let text = offered?.target.original ?? lastTranscript else { return }
+        // frozen when the offer went up, and back into the field frozen with
+        // it: the panel can be open for a while, and `focusAtPress` by then can
+        // belong to a dictation that finished in the meantime. Teaching a rule
+        // is the panel's own and is unchanged.
+        guard let target = offered?.target else { return }
         Log.write("offer: taken; opening the correction panel")
+        // Nothing here is a selection, and the offer's own target is what says
+        // where the words go. Cleared so no later save can read one left behind
+        // by an earlier flow.
         pendingSelection = nil
         correctionPanel.onSave = { [weak self] rules, correctedText in
-            self?.saveCorrections(rules, correctedText: correctedText)
+            self?.saveOfferedCorrection(rules, correctedText: correctedText, into: target)
         }
-        correctionPanel.onCancel = { [weak self] in self?.pendingSelection = nil }
-        correctionPanel.show(selection: text)
+        correctionPanel.onCancel = { Log.write("offer: correction dismissed") }
+        correctionPanel.show(selection: target.original)
     }
 
     /// This press's dictation is over, however it ended. Drops the pane it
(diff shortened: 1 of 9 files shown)


znat:
Valid, and fixed in 063e5af.

The offer already froze the element and its app in `offeredCorrection.target`
so nothing downstream would have to work out where the words go. The Correct
path then handed the panel to `saveCorrections`, which works it out anyway —
`pendingSelection`, then `focusAtPress`, then the clipboard. That threw the
frozen target away, and the panel can be open for as long as it takes to think
about a spelling, so `focusAtPress` can belong to a dictation that finished in
the meantime.

The offered path now goes through `replace(_:with:)`, which is the routine the
tapped offer already uses for exactly this: focus handed back first, a refusal
to write unless the field is still the one that was dictated into, and the
clipboard when it cannot.

- `learn(_:)` is split out of `saveCorrections` so the rule writing is shared.
  `saveCorrections` is otherwise unchanged — the spoken correction still works
  the way it did.
- `replace(_:with:)` now returns whether the field took the sentence. A saved
  rule is announced only when it did, so a rule notice never buries the
  clipboard warning.

`swift build -c release`: 20 warnings before, 20 after. SwiftLint: 32 before,
32 after.

PR #111, opened 2026-08-11
feat: the offer fades rather than vanishing

znat:
PR 7 of the HUD stack, on top of #110. Section 3 of
[docs/proposals/hud-placement-and-offer.md](docs/proposals/hud-placement-and-offer.md),
row 8 of its table.

The offer held full strength for three seconds and then disappeared. It now
starts just below full, thins out over nine seconds, holds while the pointer is
on it, and goes at once when you decide something.

## The decisions, and they are decisions

- **Below full to start.** 0.92. It is the one surface that appears without
  being asked for, so it should announce itself less than a notice does.
- **Linear, not eased.** An eased fade spends its time near the ends and
  crosses the middle fast, which reads as being yanked away halfway through. A
  straight ramp says "this is running out" at a steady rate — a clock without a
  clock.
- **Usable to the last frame.** The keys and the chips work the whole nine
  seconds. The fading only says how long is left.
- **Nine seconds, and it can afford to be**, because the pointer resets it.
  Three was not long enough to read the sentence, decide it is wrong and reach
  for a key. A fade this generous would be clutter after every dictation if
  using it did not put it back.
- **The pointer stops the clock, it does not reset it**, and it starts again
  from the beginning when the pointer leaves. A surface that went on thinning
  while you reached for it would be arguing about whether you had finished.
- **A decision cuts it at once.** Escape, Return or running a command takes the
  pill immediately rather than letting the fade play out. Dismissing is a
  decision, and a decision that took two more seconds to show would read as the
  key not working. A *running* decay is cut, because its alpha is already on
  its way to zero and a fade laid over that has nothing left to move. One the
  pointer is *holding* is stopped at 0.92, which is a real strength, so it
  leaves as an ordinary 0.18s fade — clicking a chip is the common path and the
  pointer is on the pill by definition, so that is the one that must not blink
  out.
- **Any new pill state cancels the decay and snaps back to full.** A decaying
  pill that then had something to say would otherwise say it at 20%.

## The generation number is not decoration

Restarting or cancelling a decay makes the animation it replaced fire its
completion handler immediately. A boolean cannot tell the two apart: the new
decay sets `isDecaying` back to true before the old handler runs, so the
superseded handler passes the guard and orders the panel out. That was a live
bug in the spike — the very thing meant to hold the offer open dismissed it.

`decayRun` is bumped by every path that replaces or cancels a decay, and the
completion handler does nothing unless it is still the live one. The naive
version looks correct, which is why the comment says what the number is for.

## Three clocks, moving together

The pill's own fade, the work item that ends the offer when nobody answers it,
and the tap's expiry. They have to agree, or the pill is on screen with keys
that no longer work, or the reverse.

`OfferKeys.extend(until:)` comes back for the third. The PR that added the tap
left it out as unused and said it belonged here; this is where the deadline
moves.

While the pointer is inside, the offer has no deadline at all. `offerUntil`
becomes a date that never arrives, which is why the work item is not armed with
it — `asyncAfter` at that distance is an overflow rather than a long wait.
Leaving arms a whole new nine seconds. The tap's expiry is a backstop against a
tap that outlived its offer, and a pointer resting on the pill is the opposite
of that. Every path that ends the offer still calls `OfferKeys.stop`.

**A hover that never leaves cannot be believed.** `onHover` is the only thing
saying the pill is still under the pointer, and a Space change, Mission Control
or a window ordered out from underneath it can all swallow the exit. That would
hold the offer open until the next dictation, with the letters it takes from
every app. So while it is held the work item is armed anyway, every
`offerSeconds`, and it asks `PillHUD.pointerIsOver` rather than trusting the
last event. A pointer that has gone without saying so is treated as one that
said so: the offer gets its nine seconds and runs out normally. The hold is
indefinite; the tap is not.

## The preview shows the real thing

`--panels offer` went through a bare `set(.offer(…))`, which held it at full
strength for as long as you left it. That was right when the offer had no
entrance of its own; it is wrong now, because a preview at alpha 1 is a picture
of a pill that never appears. It goes through `PillHUD.offer(_:for:)` instead,
and the hover closure calls `hovering(_:)` — so the preview fades like the real
one, and the pointer holds it open, which is how you keep it on screen to look
at. `AppDelegate.offerSeconds` stops being private for it: a preview with a
number of its own is a preview of something else.

## What was checked

- `swift build -c release` — 20 warnings, the same 20 pre-existing
  `AppDelegate` Sendable ones as on the base. No new ones.
- `swiftlint lint` — 32 warnings, the same 32 as on the base.
- Every deterministic script in `.github/workflows/checks.yml`, all green:
  numbers 97/97, replacements 23/23, pipeline 71/71, wake 24/24, split 14/14,
  dotted 54/54, dates 26/26, transform folders 12/12, eval 25/25, audio
  recovery 11/11, possessive 35/35, default config, vocabulary config, seeded
  transform 61/61, pinned certificate, no voice 5/5.
- `--check-config` still reads the config and prints the offer.

## What was not verified

**None of this has been seen running.** Every claim above is about pixels and a
pointer, and nothing in the check set has a screen or a mouse:

- That 0.92 reads as "below full" rather than as a rendering fault.
- That the linear ramp reads as a clock over nine seconds, and that nine is the
  right number rather than seven or twelve.
- That the pointer entering the pill holds it, and that leaving starts the nine
  seconds again — including that SwiftUI's `onHover` fires for both on a
  `nonactivatingPanel` that ignores the mouse in every other state.
- That a letter typed at second 8.5, when the pill is nearly transparent, still
  runs its command.
- That Escape, Return and a click all take the pill at once rather than leaving
  a ghost of it on screen — and that the cut on the running path and the fade
  on the held path both read as "gone" rather than as a glitch. The split rests
  on a claim about AppKit that was reasoned, not measured: that during
  `animator().alphaValue = 0` the property already reads zero.
- That a notice raised over a decaying offer comes up at full strength.
- That the generation guard holds in the app and not only in the reading of it.
- That an exit really can go missing, and that `pointerIsOver` answers
  correctly when it does. The screen coordinates it compares are the right ones
  on paper; nobody has watched it recover.

`scripts/check-inplace.sh` and `scripts/check-span.sh` were **not** run: they
replace `/Applications/ParrotFlowDev.app` and take the keyboard, and somebody
was dictating with that app.

## Not in this PR

When the offer is raised is PR 8. What a command does is PR 9.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -257,19 +257,35 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var offerOnScreen: [OfferedCommand]?
     /// Gives the keys back when the offer simply runs out.
     ///
-    /// The pill takes itself off screen after `offerSeconds` and tells nobody,
+    /// The pill fades itself off screen after `offerSeconds` and tells nobody,
     /// so an offer that nobody answers ends with no call back into here. This
     /// is that call. Without it the tap's own expiry would be the only thing
     /// left to end it, and that is the backstop, not the way out.
+    ///
+    /// It has a second job while the pointer is holding the offer open. See
+    /// `offerDeadlinePassed`.
     private var offerKeysExpiry: DispatchWorkItem?
+    /// True while the pointer is resting on the offer.
+    ///
+    /// The clock is stopped then, and `offerUntil` becomes a date that never
+    /// arrives — which is also why the deadline above is not armed with it:
+    /// `asyncAfter` at that distance is an overflow rather than a long wait.
+    /// See `holdTheOffer`.
+    private var offerHeld = false
 
     /// How long the offer stays up.
     ///
-    /// Long enough to read the key and act on it, short enough that it is gone
-    /// before you have started typing the next thing. It is on screen after
-    /// every dictation, so any longer and it stops being an offer and becomes
-    /// something in the way.
-    private static let offerSeconds: TimeInterval = 3
+    /// Long enough to read the sentence, decide it is wrong and reach for a
+    /// key, which three seconds was not. It can afford to be this long because
+    /// it does not sit there at full strength: it thins out the whole way — see
+    /// `PillHUD.offer` — and the pointer stops that clock and gives the nine
+    /// seconds back in full. A fade this generous would be clutter after every
+    /// dictation if reading it did not put it back.
+    ///
+    /// Not private: `--panels offer` previews the offer for as long as the app
+    /// gives it, and a preview with a number of its own is a preview of
+    /// something else.
+    static let offerSeconds: TimeInterval = 9
 
     /// What the offer offers: Correct, then every transform that asked for a
     /// place on it with `offer: true`.
@@ -2001,6 +2017,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         }
 
         offerUntil = Date().addingTimeInterval(Self.offerSeconds)
+        // A new offer is never born held, whatever the last one ended as.
+        offerHeld = false
         offerPressRun = press.run
         // This dictation's own words and its own field, frozen with the offer.
         offeredCorrection = (press.run, Correction(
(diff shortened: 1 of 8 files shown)


znat:
Fixed: `--panels offer` now goes through `PillHUD.offer(_:for:)` and wires `hovering(_:)` into its hover closure, so the preview arrives at 0.92, fades over the same nine seconds the app gives it, and is held open by the pointer. `AppDelegate.offerSeconds` is no longer private, so the preview reads the shipped number instead of carrying one of its own. `docs/cli.md` says so.

PR #113, opened 2026-08-11
feat: the offer follows every dictation, and a command rewrites in place

znat:
`insertDictation` ends five ways. Only one of them raised the offer:

| Ending | Was | Now |
|---|---|---|
| focus moved since the press | clipboard, notice on the pill | clipboard, notice in the menu bar, **offer** |
| nowhere to type | clipboard, notice on the pill | clipboard, notice in the menu bar, **offer** |
| `.pasted` | field, offer | unchanged |
| `.copied` (`insert_mode: clipboard`) | clipboard, menu bar | **offer** as well |
| `.clipboardOnly` (no Accessibility) | clipboard, menu bar | **offer** as well |

A word the recogniser got wrong is worth teaching whether or not the sentence
reached a text field, and the offer cannot wait — it is about the words you are
looking at now. The clipboard notices are worth keeping and worth less than the
offer, so the two `flash` calls become `setLabel`: the pill is the offer's, and
the menu bar is where a message that is not urgent lives. The `Log.write` lines
are untouched.

This PR also folds in spec PR #10, `a command from the offer rewrites in
place`. Both halves answer one question — when a command is taken from the
offer, where does the result go — and the answer is one mechanism.

## Where a command writes back

This is the claim worth checking, so here it is plainly. `Correction` carries a
`landing`, frozen when the offer goes up, and it decides:

- **Into the field** — after `.pasted`, and only then. The app is brought
  forward, the focused element is compared against the one that was dictated
  into, and the sentence is substituted in place. That path is unchanged apart
  from the label it logs.
- **Onto the clipboard** — after the other four. The corrected or rewritten
  sentence replaces the clipboard, so the ⌘V you were about to press pastes it.

Nothing writes into `press.element` on the four clipboard endings. That element
is nil on the "nowhere to type" path, and on the other three it is a field these
words never reached — writing into it is exactly the "a sentence in somebody
else's window" failure the focus check exists to prevent. The app is not
activated on those paths either: after a dictation that ended *because* focus
moved, pulling the app you just left back in front is the second surprise in a
row.

Two smaller consequences on the clipboard endings. The landed-diff is skipped —
nothing landed, so it would only report something else moving on screen — and
the pane snapshot is retired anyway rather than left for the sweep.

## A transform from the offer

A chip used to call `runTransform(found, instruction: "")`. That is the
*selection* path: it takes `selectionAtPress`, or reads the selection in
whatever is frontmost, and falls back to `lastTranscript`. After a clipboard
ending there is no selection belonging to this dictation, so it could rewrite
unrelated text in an unrelated window — on the "focus moved" path, in another
window by definition.

`runOfferedTransform` replaces it. It runs over the sentence the offer froze and
hands the result to the same write-back Correct uses, so the landing decides
where it goes. No preview, whatever `confirm:` says: nothing is being
overwritten, the words are on screen above the pill, and you chose the command
with a key. `runInline` already makes that argument.

It fails open. A transform that errors, returns nothing, or changes nothing
leaves the words exactly as they are — in the field or on the clipboard — and
says why in the menu bar rather than on the pill, which sits over the words the
message is about.

## Comments that became false

Three, rewritten rather than deleted: `showCorrectOffer`'s "Only after a paste
that worked" paragraph, the `defer { dictationEnded(press.run) }` note that said
every other path makes no offer, and `runOfferedCommand`'s note that a transform
still shows its preview. `docs/configuration.md` and `docs/pipelines.md` said
both of those things too.

## Checked

- `swift build` — 20 warnings, the same 20 `origin/main` has. None new.
- `swiftlint lint` — one pre-existing warning in this file (`CFEqual(now!, …)`),
  none new.
- `scripts/check-default-config.sh`, `check-pipeline.sh` (71/71),
  `check-no-voice.sh` (5/5), `check-transform-folders.sh` (12/12).
- Not run: `check-inplace.sh` and `check-span.sh` replace the installed dev app
  and take the keyboard, and this machine is being dictated into.
- Not run on a live mic. The spec asks for a dictation down each of the four
  endings; that needs the person at the keyboard.

## After review

Two write-backs were tightened in review, both about the clipboard:

- A clipboard landing carries `NSPasteboard.changeCount` from the moment the
  words were put there. Copy anything before the command finishes and the
  rewrite is not written over it (e4575e0).
- The last-resort copy — the field refused the rewrite — checks the same thing
  from the moment the command was taken (dada54b). The sentence is untouched in
  the field there, so a refusal costs only the rewrite.
- The read and the pasteboard write sit next to each other with nothing between
  them (9ecd384). `NSPasteboard` has no compare-and-write, so the gap can only
  be made small, and that is said on the function.

Three findings are rebutted in the thread rather than fixed: the fallback copy
using the command-time count (the fallback exists so a correction is never
lost), the last identical occurrence winning a rewrite (`Surface.range(of:)`'s
documented rule, and ⌘V reports no range back), and check-then-write not being
atomic (no such primitive exists). Each reason is written into the code as well
as into the thread.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -1984,13 +1984,19 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         return offerUntil > Date()
     }
 
-    /// Say what can be done about the words that just landed, and for how long.
+    /// Say what can be done about the words just dictated, and for how long.
     ///
-    /// Only after a paste that worked. Every other ending already has something
-    /// to say — "on your clipboard", "grant Accessibility" — and those are more
-    /// useful than an offer, so they keep the pill. This is the one ending that
-    /// had nothing on it at all: the pill simply vanished and the dictation was
-    /// over, which is a fine thing to feel and a wasted second of screen.
+    /// After every ending, not only a paste that worked. A word the recogniser
+    /// got wrong is worth teaching whether or not the sentence reached a text
+    /// field, and the offer cannot wait: it is about the words you are looking
+    /// at now. The messages the clipboard endings used to put on the pill —
+    /// "on your clipboard", "grant Accessibility" — are worth keeping and worth
+    /// less than the offer, so they moved to the menu bar, where messages that
+    /// are not urgent live. See `insertDictation`.
+    ///
+    /// `landing` is where those words actually went. It is frozen into the
+    /// correction, and it is what keeps a correction out of a field the
+    /// dictation never wrote into.
     ///
     /// The commands are read from the config, so what is on the pill is what
     /// that machine can actually do. It no longer waits on the hotkey being
@@ -2000,7 +2006,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// `press` is the dictation this offer is about, carried down from its own
     /// key-down — see `insertDictation`. Nothing below reads press-time state
     /// off `self`, so an offer can never be moved by another dictation's press.
-    private func showCorrectOffer(for press: Press) {
+    private func showCorrectOffer(for press: Press, landing: Correction.Landing) {
         guard config.feedback.correctOffer else { return }
         guard let text = lastTranscript?.trimmingCharacters(in: .whitespacesAndNewlines),
               !text.isEmpty else { return }
@@ -2022,7 +2028,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         offerPressRun = press.run
         // This dictation's own words and its own field, frozen with the offer.
         offeredCorrection = (press.run, Correction(
-            original: text, element: press.element, owner: press.owner
+            original: text, element: press.element, owner: press.owner, landing: landing
         ))
         let commands = offerCommands
         pill.offer(commands, for: Self.offerSeconds)
@@ -2053,8 +2059,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // by run, so it is this dictation's own pane and nobody else's. A
         // remembered anchor is still a guess, and this is what replaces it with
         // the answer.
-        if let pane = screenAtPress.removeValue(forKey: press.run)?.text,
-           let element = press.element {
+        //
+        // Dropped on every ending, not only the one that uses it. A pane
+        // belongs to one run and no later dictation can want it, so it is
+        // retired here as `dictationEnded` would retire it — leaving it for the
+        // sweep would be keeping a copy of somebody's screen for no reason.
+        let pane = screenAtPress.removeValue(forKey: press.run)?.text
+        // Nothing landed in a field on the clipboard endings, so the diff has
+        // nothing to find. Whatever it did find would be something else moving
+        // on screen.
+        if landing == .field, let pane, let element = press.element {
             findWhereTheWordsLanded(comparedWith: pane, in: element, for: press.run)
         }
     }
@@ -2214,6 +2228,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         endTheOffer()
         pill.hide()
 
+        // The sentence and where it went, frozen when the offer went up. Both
+        // commands need it, and neither has anything to work on without it.
+        guard let target = offered?.target else { return }
+
         // Looked up again rather than carried on the chip, so a transform
         // deleted from a config reloaded while the offer was up is not run
         // from a chip that outlived it.
@@ -2223,10 +2241,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 return
             }
             Log.write("offer: taken; running \"\(name)\"")
-            // No instruction: the chip is the whole of what was asked for.
-            // This still shows the transform's preview, because that is what
-            // `confirm:` asks for today.
-            runTransform(found, instruction: "")
+            runOfferedTransform(found, over: target)
             return
         }
 
@@ -2239,7 +2254,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // it: the panel can be open for a while, and `focusAtPress` by then can
         // belong to a dictation that finished in the meantime. Teaching a rule
         // is the panel's own and is unchanged.
-        guard let target = offered?.target else { return }
         Log.write("offer: taken; opening the correction panel")
         // Nothing here is a selection, and the offer's own target is what says
         // where the words go. Cleared so no later save can read one left behind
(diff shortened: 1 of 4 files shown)


znat:
Both findings read. One fixed, one rebutted.

**1. Late clipboard write-back overwrites a newer copy — fixed** in e4575e0.

`Landing.clipboard` now carries `NSPasteboard.changeCount` as it was straight
after the words were put there (`Landing.clipboardNow()`, called at each of the
four clipboard endings). `replace` refuses to write when the count has moved:

```
guard NSPasteboard.general.changeCount == change else {
    Log.write("offer: the clipboard has been used since the dictation; …")
    flash("Clipboard has changed — \(what) not copied", tone: .caution)
    return false
}
```

The window is real and longer than the transform: the correction panel can be
open for as long as it takes to think about a spelling. Losing a rewrite costs a
second attempt; taking away something this app never put there costs whatever
that was. Documented in `docs/configuration.md`.

**2. A later identical occurrence gets rewritten — not fixing, and I think
`.backwards` is the right rule here.**

Three reasons.

It is not this PR's behaviour. `replace` -> `applyInPlace(fuzzy: false)` is the
path Correct has taken since #109, and the correction panel holds it open far
longer than a transform does. This PR adds a second caller to it and changes
nothing about how it locates text.

The last occurrence is the right one for the offer. The offer is always about
the newest dictation, and the newest dictation is the last thing written into
the field. Dictating the same short sentence twice — "Yes.", "Done." — is
ordinary, and `.backwards` then rewrites the one the offer is about. The failure
case needs the user to type a byte-identical sentence into the same field during
the seconds a transform is running. Refusing on ambiguity would trade a common
case for a rare one.

The proposed fix belongs elsewhere. A durable range or a bounded-context match
is a change to `applyInPlace`, which `saveCorrections` and `applyTransform`
also use. It would change what those two do, and it should be argued with them
in view rather than added here.

Worth noting that `replaceSelected` already refuses on ambiguity — "N copies of
the selection and nothing says which was meant; not guessing" — because there
the text was pointed at with a mouse and nothing says which. Here the text was
dictated a moment ago, and *last* is what says which.

znat:
Second round.

**Field fallback overwrites a newer clipboard — fixed** in dada54b.

It is the same rule as the clipboard landing, applied to the last-resort copy.
`runOfferedCommand` reads `NSPasteboard.changeCount` at the moment the command
is taken — that is the start of the wait that makes any of this possible — and
both write-backs are handed it. The fallback now refuses when the count has
moved.

The refusal is cheaper here than on a clipboard landing: the field would not
take the rewrite, so the sentence is still sitting there untouched. What a
refusal costs is the rewrite, and a second attempt gets it back.

Note the change count is read when the command is taken, not when the offer goes
up. `TextInserter` restores the previous clipboard 0.4s after a paste, so a
count taken at the offer would be stale for every field landing and the fallback
would refuse every time.

**The later identical occurrence — still not fixing.** The reasons in my
previous comment stand, and the second review moved the finding onto
`applyInPlace:1720`, which is where I said it lives. To be concrete about what a
fix needs:

Telling "the copy I wrote" from "an identical copy you typed afterwards" cannot
be done from the text, because both are the last occurrence. It needs the range
the insertion actually occupied, and nothing captures one: `TextInserter` posts
⌘V and the target app takes the pasteboard when it gets round to it, so this
process never learns where the words went. (The proposal calls that out — §6,
"a pane read after the paste is not a baseline, and nothing here can tell".)

Refusing on ambiguity instead would break the ordinary case. Dictating the same
short sentence twice is normal, and `.backwards` picks the one the offer is
about — the newest dictation is the last thing written.

So the fix is "the insertion reports where it landed", it changes `applyInPlace`
for `saveCorrections` and `applyTransform` as well, and it belongs in a PR that
can argue it with those two in view. Not in this one.

znat:
Third round. The one finding left is the later identical occurrence, and I am
holding the rebuttal. 20b6a2f writes the reasoning onto the call site rather
than leaving it to be rediscovered.

Three things, in order of weight.

**The rule is the repo's, and it is deliberate.** `Surface.range(of:)`:

> Last rather than first: the word being corrected was dictated a moment ago, so
> the most recent occurrence is the one meant.

The offer is always about the newest dictation, and the newest dictation is the
last thing written into the field. Say "Yes." twice and the second one is the
one the offer is about, and the one `.backwards` corrects.

**The proposed fix is not available.** "Capture the dictated insertion span when
the dictation lands" needs the span, and this process never learns it.
`TextInserter` posts ⌘V; the target app reads the pasteboard when it gets round
to it and reports nothing back. The proposal file says the same thing in §6 —
"a pane read after the paste is not a baseline, and nothing here can tell". A
real fix is the insertion reporting where it landed, which is a change to
`TextInserter` and `Surface` ahead of this work rather than inside it.

**Refusing on ambiguity would cost more than it saves.** The alternative offered
is "refuse when duplicate occurrences cannot be disambiguated". A terminal
echoes what you send it, so a second copy of a dictated line appears there
routinely and for reasons that have nothing to do with the user. Terminals are
most of what this stack was built for — see §1 of the proposal. Trading a
reliable correction in a terminal for a rare case needing a byte-identical
sentence typed into the same field during the seconds a transform runs is the
wrong way round.

`replaceSelected` does refuse on ambiguity, and the difference is the point: it
is resolving text somebody pointed at with a mouse, where nothing says which
copy was meant. Here the text was dictated a moment ago, and *last* is what says
which.

Both of the findings that were about this PR's own new writes are fixed: the
clipboard landing in e4575e0, the field fallback in dada54b.

znat:
Fourth round.

**The clipboard check-then-write gap — narrowed** in 9ecd384, and I agree it
cannot be closed.

There was a `Log.write` between the guard and `clearContents()`. That is a file
write sitting inside the gap, and it was the only thing in it worth removing.
`copyOverOurOwn(_:unlessChangedFrom:)` is now the read and the two pasteboard
calls with nothing between them; every message moved out to the callers so
nothing can get back in.

Beyond that there is nothing to have, and the comment says so on the function.
`NSPasteboard` offers no compare-and-write. `declareTypes(owner:)` names an
owner for lazy-data callbacks; it does not stop another process writing. So
every clipboard write on this platform is a look followed by a write, including
the one `TextInserter` has done on every dictation since the app existed, and
including the one in every app on the machine. Declining to write when
exclusivity cannot be guaranteed would mean never writing the clipboard at all,
which is the feature.

**The later identical sentence — unchanged, and the rebuttal stands.** Third
time: the rule is `Surface.range(of:)`'s and it is deliberate, the span it asks
me to capture cannot be captured because ⌘V reports nothing back, and refusing
on duplicates would break corrections in a terminal, which echoes what you send
it. 20b6a2f writes all three onto the call site.

Both findings that were about writes this PR introduced are fixed: the clipboard
landing in e4575e0, the field fallback in dada54b.

znat:
Fifth round. Nothing changed in the code this time, and I will say why for each
of the three.

**1. The field fallback writes the clipboard even if you copied during the
offer.** Correct, and deliberate. This is the fallback saying what it has always
said: the field would not take the rewrite, so it goes on the clipboard rather
than being lost. That path predates this PR — `replace` has cleared and written
the pasteboard unconditionally since the offer was built. What this PR added is
the guard from the moment you take the command, which is the window the wait
opens.

Going further means refusing the fallback for something copied *before* you
chose. Then a correction typed into the panel, that the field then refuses,
is gone: no field, no clipboard, and nothing to try again with except saying
the sentence over. The clipboard fallback exists so a correction is never lost,
and I am not willing to trade that against a clipboard the user filled and then
chose to run a command anyway.

There is also no baseline to compare against. `TextInserter` puts the previous
clipboard back 0.4s after a paste, so a count taken when the offer goes up is
stale for every field landing, and the fallback would refuse every time.
Establishing ownership properly means the insert reporting when the pasteboard
settled, which is a change to `TextInserter` that three other callers share.

**2. The later identical sentence.** Fourth time. `Surface.range(of:)` decides
this and its reason is written there: the offer is about the newest dictation
and the newest dictation is the last thing written. The span you ask me to
capture cannot be captured — ⌘V reports nothing back. Refusing on duplicates
breaks corrections in a terminal, which echoes what you send it. All three are
now written on the call site in 20b6a2f.

**3. Check-then-write is not atomic.** Agreed, and unfixable — as the finding
itself says. 9ecd384 removed the file write that was sitting inside the gap and
put the read and the two pasteboard calls next to each other, which is as small
as the window gets. `NSPasteboard` has no compare-and-write, and
`declareTypes(owner:)` stops nobody. "Decline when exclusivity cannot be
guaranteed" is "never write the clipboard", which is the feature.

Summary of the five rounds: two findings were about writes this PR introduced
and both are fixed — the clipboard landing (e4575e0) and the field fallback
guard (dada54b). The three above are a pre-existing designed fallback, a
documented rule in shared code, and a platform limitation. I have written each
reason into the code rather than only into this thread.

znat:
The third finding is tracked as #114 rather than fixed here.

The baseline cannot be taken at the dictation without moving the snapshot into `TextInserter`'s 0.4s restore completion — read any earlier and the app's own restore moves `changeCount` before the offer is answered, so the guard would decline every time and the fallback would never fire. That is real machinery and it is not what this PR is about.

The trade as it stands: declining costs a correction typed by hand with nothing to retry from, overwriting costs a clipboard item and says so on screen. Accepted deliberately, argued at the call site.

The other two findings stand rebutted. `NSPasteboard` has no compare-and-write, so no check on it can be atomic; and `Surface.range(of:)` taking the last copy of a repeated sentence is pre-existing and cannot be fixed from here, because ⌘V reports no insertion range back.

PR #115, opened 2026-08-11
feat: a Bluetooth microphone says so, once

znat:
Section 5 and PR 11 of `docs/proposals/hud-placement-and-offer.md`. Stacked on
#113 — the diff below is against `feat/the-offer-follows-every-dictation`.

## What it does

AirPods and most headsets record over a Bluetooth voice profile. It narrows the
band, gates quiet sound and adds delay, so the ends of your words go missing and
nothing on screen says why. The app now says it — once, for as long as you stay
on that microphone, just after a dictation.

| | |
|---|---|
| `Recorder.InputDevice` | UID, name and transport of the device the engine is bound to |
| `MicNotice` | The panel, bottom right, and the one slot behind it |
| `Press.mic` | The microphone frozen when the recording starts |
| `showCorrectOffer` | One line, at the end of every dictation |
| `--panels microphone`, `--panel-sheet` | The surface on its own, and both its states |

## Three shapes this is not

Each was built and thrown away. The file's doc comment keeps the reasons.

**Not a line on the pill.** It was either in the way of the offer or riding
under the volume meter, saying the same sentence into every dictation.

**Not a warning every time.** This is not a status. It is a thing to read,
decide about, and not be told again.

**Not the three reasons up front.** They are true and worth having, and they are
not what you need at the moment it appears — what you need then is whether to go
and change the microphone. So they sit under a *Why*.

## Once, and what "once" is measured against

`lastDevice` is one slot: the microphone the last dictation was recorded on, by
UID. Not a list of the ones already mentioned — that is the difference between "once ever"
and what this is meant to be. Changing microphone forgets the one you left, so
dictating on another and coming back is a decision being revisited, and it is
said again.

The slot is set for a wired microphone too. It holds which microphone was last
decided about, not which one was last complained about, and a dictation on the
built-in mic is what makes going back to the headset a change.

It only sees the microphone a dictation was on. Switching away and back with
nothing dictated in between is not a change it can see.

## The microphone that recorded the words

Read in `startRecording`, right after `recorder.start` opens the device, and
frozen onto the `Press` beside the element and the app. Transcription is async:
read at the end instead, a headset that disconnected in that second takes the
notice with it, and one that connected gets named for a dictation it never
heard. That struct exists so nothing about a dictation is read off `self` after
the fact.

Taken from the engine's own binding, not from the system default. `bound` is
the device the running engine was prepared against; the default can move the
line after `start` returns, and a headset that connects in that gap would get
named for a dictation it never heard. `Recorder.boundDevice` reads that field
once and asks that one device ID for its UID, its name and its transport — two
lookups of "what is default now" can straddle a change and describe two
devices.

The UID is what decides, and the name is only what it says out loud. Two
microphones can answer to "Headset Microphone", and with the name as the key the
wired one silences the notice for the Bluetooth one. Where a device gives no
UID, its CoreAudio device ID stands in — unique among the devices attached at
once, and not stable across a reconnection, which costs a notice repeated rather
than a notice missed.

## The transport, not the name

A list of brand names is wrong the day somebody buys a headset nobody thought
of, and it was the wrong question anyway. What costs you the ends of your words
is the transport: a Bluetooth voice profile narrows the band and gates quiet
sound whatever is printed on the case.

## Where it is raised

`showCorrectOffer(for:landing:)`, which all five endings of `insertDictation`
reach — pasted, copied, no Accessibility, focus moved, nowhere to type. On
every dictation, not only the ones that paste cleanly.

It sits **above** that function's guards, and that is the one place this differs
from the spike. Those guards are about the offer, not about the microphone:
`correct_offer: false` is a choice about chips on the pill, an empty transcript
is the symptom itself, and a dictation that lost the pill to a newer one was
still recorded on the same microphone.

It never takes focus. It arrives when you are already typing again, and a dialog
that stole the keyboard to give advice would be a worse citizen than the
microphone it is complaining about.

## What it looks like

Both states are on `--panel-sheet`, with a device name long enough to say
whether it fits.

`--panels microphone` puts the real panel on screen with a made-up name —
`show(mic:)` rather than `showIfNeeded`, so it appears on a machine whose
microphone is wired. *Why* and *Got it* both work there.

## Checks

- `swift build`: **20 warnings before, 20 after** — the same 20, all
  pre-existing Swift 6 concurrency warnings in `AppDelegate.swift`, moved by a
  line or two.
- `swiftlint lint`: **32 warnings before, 32 after**, identical set, 0 errors.
- `scripts/check-audio-recovery.sh` 11/11, `check-default-config.sh` ✓,
  `check-no-voice.sh` 5/5. The rest of the check scripts score prompts and
  patterns this PR does not touch. `check-inplace.sh` and `check-span.sh` were
  not run: they replace the installed app and take the keyboard.
- `--panel-sheet` and `--panels microphone` both render.

## Docs

`docs/configuration.md` gains **The microphone** — where the input device comes
from, what Bluetooth costs, and what "once" means. `docs/cli.md` gains the new
`--panels` case.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -68,6 +68,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var labelToken = 0
     private let correctionPanel = CorrectionPanel()
     private let previewPanel = PreviewPanel()
+    /// Says once per microphone that this one will cost you words.
+    private let micNotice = MicNotice()
     private var pendingSelection: SelectionReader.Selection?
     /// Captured the moment the hotkey goes down — see SelectionReader.snapshot.
     private var selectionAtPress: SelectionReader.Selection?
@@ -2007,6 +2009,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// key-down — see `insertDictation`. Nothing below reads press-time state
     /// off `self`, so an offer can never be moved by another dictation's press.
     private func showCorrectOffer(for press: Press, landing: Correction.Landing) {
+        // Beside the offer, not on it: its own window, so advice about the
+        // microphone never costs you the chance to fix the sentence. Here
+        // because this is the end of a dictation and every ending comes
+        // through — a Bluetooth mic is worth saying something about the moment
+        // you have watched a transcript come back short, and worth nothing at
+        // launch, when you were not dictating.
+        //
+        // Above the guards below, which are all about the offer rather than
+        // about the microphone: `correct_offer: false` is a choice about
+        // commands on the pill, an empty transcript is the symptom itself, and
+        // a dictation that lost the pill to a newer one still used the same
+        // microphone. It is said once per microphone, so none of them can make
+        // it say it twice either.
+        micNotice.showIfNeeded()
+
         guard config.feedback.correctOffer else { return }
         guard let text = lastTranscript?.trimmingCharacters(in: .whitespacesAndNewlines),
               !text.isEmpty else { return }
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -82,12 +82,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// without Accessibility — an app condition has no business needing a
     /// permission that gating a stage by app does not otherwise require.
     private var appAtPress: Pipeline.App?
-    /// Which microphone the engine opened for this recording, and whether it
-    /// is on Bluetooth. Read when recording starts and frozen onto the `Press`
-    /// when the clip is handed to the decoder, so the microphone notice names
-    /// the device that recorded the words rather than whatever is default a
-    /// second later. One slot is enough: only one recording runs at a time,
-    /// and it is taken the moment that recording stops.
+    /// Which microphone the engine is recording through, and whether it is on
+    /// Bluetooth. Read when recording starts and frozen onto the `Press` when
+    /// the clip is handed to the decoder, so the microphone notice names the
+    /// device that recorded the words rather than whatever is default a second
+    /// later. One slot is enough: only one recording runs at a time, and it is
+    /// taken the moment that recording stops.
     private var micAtPress: (name: String, isBluetooth: Bool)?
     /// That same app's icon, for the pill. Held apart from `appAtPress` because
     /// `Pipeline.App` is what the pipeline matches on and has no business
@@ -955,10 +955,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
         do {
             try recorder.start(config: config)
-            // Asked here, where the engine has just opened the device, and not
-            // when the transcript comes back. One lookup, so the name and the
-            // transport are the same device's — see `Recorder.inputDevice`.
-            micAtPress = Recorder.inputDevice
+            // The engine's own device, asked here rather than when the
+            // transcript comes back. Not the system default: that is a
+            // different question the moment a headset connects, and the notice
+            // is about the microphone these words went through. See
+            // `Recorder.boundDevice`.
+            micAtPress = recorder.boundDevice
         } catch {
             // A notice, not an alert. `runModal` holds the main run loop, and
             // the hotkey is delivered on it: one failed press behind a modal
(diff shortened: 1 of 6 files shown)


znat:
Both findings were right, and both are fixed in 4f766a3.

**1 — returning to a microphone said nothing.** The `told` set was never
emptied, so the file's own comment ("changing to another and back says it
again") described behaviour the code did not have. It is one slot now,
`lastMic`: the microphone the last dictation was recorded on. Changing
microphone forgets the one you left, so coming back to it is a new decision and
is said again. The slot is set for a wired microphone too — it is which
microphone was last decided about, not which one was last complained about, and
a dictation on the built-in mic is what makes going back to the headset a
change.

One limit, written down rather than fixed: it only sees the microphone a
dictation was on. Switch away and back with nothing dictated in between and
there is no change for it to see. Watching CoreAudio for device changes would
close that, and it would be a listener running all day to catch a case where
nothing was said in between anyway.

**2 — it read the microphone a second too late.** Correct, and the same race
the rest of this function already avoids. The microphone is now read in
`startRecording`, right after `recorder.start` opens the device, and frozen onto
the `Press` with the element and the app — that struct exists so nothing about a
dictation is read off `self` after the fact. `showIfNeeded(mic:isBluetooth:)`
takes both, and asks CoreAudio nothing.

Not done: keying on the device ID rather than the name. Two devices with the
same name would share one slot, and the cost of that collision is one notice
missed on a machine with two identically named microphones. The name is what
the notice says out loud and what the menu bar shows, so it is also the thing a
reader can check.

znat:
Right again, and fixed in aba30a6.

`inputDeviceName` and `inputIsBluetooth` each asked CoreAudio which device is
default, so the pair could describe two devices — the window is small, and the
event that closes it is exactly the one this notice is about: a headset
connecting and macOS making it the input.

`Recorder.inputDevice` now takes the device ID once and asks that ID for both
the name and the transport. `inputDeviceName` keeps its own lookup for the menu
bar and `--audio-recovery`, which print a name and decide nothing;
`isBluetooth(_:)` is private and takes a device ID, so there is no way left to
ask about the transport of a device you did not name.

Build: 20 warnings before this PR, 20 after, the same 20. `swiftlint`: 32
before, 32 after, identical. `scripts/check-audio-recovery.sh` 11/11.

znat:
Fixed in c76b65c, and the fix removes the window rather than narrowing it.

`Recorder.inputDevice` asked CoreAudio what the default input was, one line
after `recorder.start` returned. You are right that a headset connecting in that
gap becomes the default and would get named for a dictation it never heard.

There is a device that cannot be wrong here, and the recorder already keeps it:
`bound`, the binding the running engine was prepared against. That is the device
the words are actually going through — `start` compares it against the system
default and rebuilds the engine when they differ, so after `start` returns it is
what is recording. `Recorder.boundDevice` reads that field under `stateLock` and
asks that device ID for its name and its transport. Nothing asks what the
default is any more.

`scripts/check-audio-recovery.sh` still 11/11 — that is the set for a device
moving underneath the engine. Build 20 warnings before and after; `swiftlint` 32
before and after, identical.

znat:
Fixed in 940c35e. I had written this one off as a collision nobody would hit;
the example makes it a real one, so it is keyed properly now.

`Recorder.InputDevice` carries CoreAudio's device UID beside the name and the
transport. `MicNotice` compares UIDs; the name is what it says out loud and
nothing else. So a wired "Headset Microphone" no longer silences a Bluetooth one
with the same name, and the UID survives unplugging and reconnecting, which
`AudioDeviceID` does not.

Where a device gives no UID the name stands in. That is a device refusing to
identify itself, which is rarer than two devices sharing a name, and the string
is only ever compared with another read of itself.

Build 20 warnings before and after. `swiftlint` 32 before and after, identical
set. `scripts/check-audio-recovery.sh` 11/11. `--panel-sheet` still renders both
states of the notice.

znat:
Fixed in 782883e, and you are right that the fallback undid the fix in the one
path that needed it.

It falls back to `"device-id:\(deviceID)"` now, never to the name. Two devices
attached at once always have different CoreAudio IDs, so same-named microphones
are still told apart when neither offers a UID. What the ID cannot do is survive
a reconnection — macOS hands it back out — so the same headset can be mentioned
a second time. That is the direction to be wrong in: this is a warning you would
rather hear twice than not hear.

Build 20 warnings before this PR and 20 after. `swiftlint` 32 before and after,
identical set. `scripts/check-audio-recovery.sh` 11/11.

PR #117, opened 2026-08-11
feat: the correction panel is the sentence you said, edited in place

znat:
PR 12 of the stack in [docs/proposals/hud-placement-and-offer.md](../blob/main/docs/proposals/hud-placement-and-offer.md) section 7.

## What the panel is now

It was a table: one row per word, "heard as" beside "should be", with text
fields. It is now the sentence you just dictated, laid out horizontally and
editable in place. What was heard sits above the word it became, struck
through at the same size. The bar under each word carries its state — faint
for untouched, sky for where you are, leaf for what will change.

    VOCABULARY

    Fix what I misheard.

                     Tasmin           Mick
     I    work  with Tasmeen and      Mik
     ──   ────  ──── ───────  ───     ───
     space splits a word in two   ⌫ at the start joins it to the word before
     ▸ Not sure what to correct?

           Undo ⌘Z   Discard   Add Tasmeen and Mik to the vocabulary ↩

The caret starts in the first word, so it reads as a row of fields rather
than a label. Every gesture is one a text field already has: type over a
word, `space` splits one word into two, `⌫` at the start joins it to the word
before, clearing a word drops it. `⌘Z` undoes the whole arrangement, `↩`
saves, escape discards. There is deliberately nothing of its own to learn.

The panel is as wide as the sentence needs, up to the screen less a margin,
so a short correction is not given a line break it does not need. Past three
lines it scrolls instead of growing.

## Why a rule is keyed on a span

A span owns two lists — the words that were **heard** and the words that
**replace** them. One heard word can become two, and two can become one,
without either side losing track of the other.

That is what a row-per-word panel could not express. `Redcrawl` decoded as
"red crawl" had no row to live on, and typing the name into the first row
taught `red => Redcrawl`, which then fired on the colour. Joining the two
words here teaches `red crawl => Redcrawl` and leaves the colour alone.

Everything below this panel already worked in spans: the fuzzy pass scans
two-word windows, `Vocabulary.widerSpans` offers the wider span for a name
the decoder split, and `autoApplies` has a branch for exactly this shape. The
panel was the last place still counting in words.

## What is not in it

**The button says "add to the vocabulary" and the code adds a replacement
rule.** `onSave` still goes to `learn()` → `ConfigWriter.addReplacement`,
which writes a rule to `config.yaml`. Nothing in the app writes
`vocabulary.yaml` today. The label is one PR ahead of the behaviour on
purpose; spec 13 closes the gap. Flagged so a reviewer does not have to find
it.

## Second pass

The panel was built, installed and tried on a real dictation, and the review
came back "it works, but it is not clear the text is editable". The second
commit is that round: the caret, the fitted width, a head that only names the
window, the instruction moved down to sit over the words, a footer of three
buttons with no status line, the confirm button in leaf naming the words it
will add, and a shut disclosure over which words are worth correcting at all.

`ActionButton` gained a glass style, a rich label and a disabled state.
`Parrot.action` and `PanelActions` are untouched — the preview panel and the
permissions window look exactly as they did.

## How it was checked

- `--panel-sheet` after every change; the correction panel is on it three
  times now, including with the disclosure open.
- The gesture set driven directly against `SpansModel` — split, join within a
  span, join across spans, clear, the arrow edges, tab, undo, punctuation
  surviving into the sentence, and the two-word-heard rule round trip.
- 20 pre-existing warnings in `AppDelegate.swift`, no new ones.

Not checked: anything needing a pointer or a keystroke on a real screen. This
machine has no Screen Recording permission for a screenshot, and the user was
dictating.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -147,6 +147,13 @@ jobs:
       - name: A possessive across a substitution
         run: scripts/check-possessive.sh
 
+      # The model behind the correction panel: the whitespace of the selection
+      # it opened on, and where the caret lands. Both are invisible in the
+      # panel, and both were wrong. Compiles the one file on its own, so it
+      # needs no window.
+      - name: The correction panel's spans
+        run: scripts/check-spans.sh
+
       # Not a validation set: it checks that the file a new install is written
       # with still teaches everything config.example.yaml documents. That has
       # drifted twice, both times silently.
--- a/Sources/ParrotFlow/CorrectionPanel.swift
+++ b/Sources/ParrotFlow/CorrectionPanel.swift
@@ -414,12 +414,14 @@ struct CorrectionView: View {
 
             // Leaf, the colour of a changed word's bar. The same promise: this
             // is the thing that is going to happen.
-            ActionButton(title: confirmTitle, styled: confirmLabel, key: "↩",
+            ActionButton(title: confirmTitle, styled: confirmLabel, key: "⌘↩",
                          filled: false, glass: Parrot.leaf,
                          enabled: model.hasChanges) { model.onSubmit?() }
-                // Plain Return, because Return already submits from inside a
-                // word — there is nowhere in this panel it would insert a line.
-                .keyboardShortcut(.return, modifiers: [])
+                // ⌘↩ is the one the button advertises, because it works from
+                // anywhere on the panel. Plain Return goes on saving too — it
+                // is handled inside the word field, where there is nowhere in
+                // this panel a newline could go instead.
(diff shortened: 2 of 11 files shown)


znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

znat:
@greptileai review

PR #121, opened 2026-08-12
feat: make text and audio logging configurable, audio off by default

znat:
## Summary

- Adds `logging.text` (default `true`) and `logging.audio` (default `false`) to `config.yaml`.
- Text logging gates `Log.write`'s file output — `~/Library/Logs/ParrotFlow.log` / `ParrotFlow-Dev.log`. NSLog (the unified system log) is untouched; the setting only controls the file, matching how the task described it.
- Audio logging gates whether a dictation's `.wav` clip survives on disk once nothing needs it any more. Capture still writes the file during recording — transcription reads it from disk — but it is now discarded: right after transcription finishes, immediately when transcription is disabled, and for a clip kept "for diagnosis" when too much audio was lost to a device change. This is a default-off behaviour change: existing users who never set `logging:` will stop accumulating clips.
- `--check-config` reports both settings, next to `feedback`.
- `config.example.yaml` and `docs/configuration.md` are updated to match. `README.md` and `docs/setup.md` are untouched, as requested.

## Decisions on the CLI commands that write to the log

- `--peek` and `--edit-test`/`--span-test` always write to the log regardless of `logging.text`. `--peek`'s own doc comment says the log is its only transport — it's meant to be launched via `open -na ParrotFlowDev --args --peek` to carry the app's Accessibility grant, and LaunchServices throws away that process's stdout. `--edit-test`/`--span-test` are launched the same way for the same reason, and `scripts/check-inplace.sh`/`check-span.sh` read their output back out of the log file (confirmed by grepping those scripts). Both commands now pass `force: true` to `Log.write`.
- Every other CLI command (`--record`, `--transcribe`, `--boost-eval`, `--spot`, etc.) just runs from an ordinary terminal, so its own `print()` output is available regardless of the log setting; their `Log.write` calls are left to respect `logging.text` like the rest of the app.
- Deliberate audio-writing commands (`--record`, `--transcribe <file>`, `--spot`, `--audio-recovery`) are unaffected by `logging.audio` — a person running them is asking for the file, and they manage their own clips directly rather than going through the app's live-dictation cleanup path.

## Test plan

- [x] `swift build` and `swift build -c release` both succeed.
- [x] `scripts/check-audio-recovery.sh` — 11/11 (its scratch config has no `logging:` key, so it now defaults to `audio: false`; verified this doesn't change any of its assertions, since it manages its own clip files directly).
- [x] `scripts/check-default-config.sh` — passes; `logging:` now matches between `Config.defaultYAML` and `config.example.yaml`.
- [x] Verified with a scratch `PARROTFLOW_CONFIG_DIR`: a config with no `logging:` key reports `text=on audio=off`; a config with only `logging: { audio: true }` reports `text=on audio=on` (partial override keeps the other default).
- [x] Ran the rest of the fast `scripts/check-*.sh` suite locally — all green (one `check-replacements.sh` run showed a known-flaky NSSpellChecker timeout, reproduced as green on rerun, and is unrelated to this change).
- [x] `swiftlint lint` — no new violations introduced.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -1023,12 +1023,15 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         guard recording || transcribing else { return }
 
         if recording {
-            // The clip on disk is left alone. It cost nothing to write, the
-            // recordings folder is where you go to find out what the app heard,
-            // and deleting the evidence of the thing you just cancelled is the
-            // opposite of useful when the reason you cancelled it was that
-            // something sounded wrong.
-            _ = recorder.stop(config: config)
+            // The clip on disk is left alone, unless `logging.audio` says
+            // recordings do not accumulate at all. Kept, it cost nothing to
+            // write, the recordings folder is where you go to find out what
+            // the app heard, and deleting the evidence of the thing you just
+            // cancelled is the opposite of useful when the reason you
+            // cancelled it was that something sounded wrong.
+            if let clip = recorder.stop(config: config) {
+                Recorder.discardIfNotKept(clip, config: config)
+            }
             // No words are coming, so this press's dictation is over here. The
             // recorder is stopped directly rather than through
             // `stopRecording`, so nothing else would retire it — and the pane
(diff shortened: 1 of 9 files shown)


znat:
Fixed in 2eb3c52: `makeOutputURL` now appends 8 hex characters from a UUID to every filename, so two captures started in the same second never collide. The root cause (second-precision timestamps) was pre-existing, but this PR is what turned it from a silent overwrite into a silent transcription failure via the new delete-once-done cleanup, so it's fixed here. Confirmed `scripts/check-audio-recovery.sh` still passes 11/11.

PR #134, opened 2026-08-13
feat: the final permissions screen invites you to dictate, live

znat:
## Summary
Follow-up to #133 (merged) — this commit landed just after that PR's squash-merge, so it never made it into main. Same content, opened as its own PR.

- The "Ready" screen now includes a third status line, "Speech model", fed live from the transcriber's real download/ready state.
- Two states: "Ready" once the model is actually ready, "Almost ready" while it's still downloading (with a live percent) — so nobody dictates into a download that hasn't finished.
- The hotkey is drawn as its own bordered mark instead of sitting as plain text in a sentence.
- Replaced the accessibility step's illustration with a native SwiftUI reproduction of the real System Settings row (icon, name, animated toggle) — reproducing what you'll actually see, not a metaphor for it.

## Test plan
- [x] `swift build` succeeds, no new warnings
- [x] Verified live: rebuilt and reinstalled to `/Applications`, both states (ready / still downloading) rendered correctly via `--panel-sheet`
- [x] Confirmed the diff against origin/main is exactly this commit's content (the other 9 commits on the old branch were already merged via #133 before this one was made)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -66,9 +66,10 @@ release machinery and gets no warning. `make hooks` refuses it locally — see
 
 ## What not to change without being asked
 
-- `Config.defaultYAML` and `config.example.yaml` are what a new install gets,
-  and several check scripts read the real file rather than a fixture. Changing
-  a default changes what everyone gets on first launch.
+- `config.example.yaml` is what a new install gets — `Config.defaultYAML`
+  reads it directly, substituting only the few lines that differ per variant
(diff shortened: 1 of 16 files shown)


znat:
Closing in favor of a clean branch — this one carried the full history of the already-merged #133 due to a squash-merge divergence, making the diff noisy (16 files instead of 3). Same content, opened as #135.

PR #143, opened 2026-08-18
feat: a spelling lesson keeps the word it is teaching

znat:
## The bug

You say "urza spells mirza" to teach a mapping. The vocabulary pass rewrites the word before `spells` into the term it already knows, so the judge sees "Mirza spells mirza". The lesson is destroyed.

## The fix

`VocabularyJudge.teaching(in:changes:)` marks any substitution whose span is followed by the word `spells`. The judge reverts those without asking a model.

Measured on the four lessons in `tests/judge-cases.yaml`:

| arm | score |
| --- | --- |
| the rule | 4/4 |
| `gemma4:e4b-mlx` | 0/4 |
| `gemma3:4b` | 0/4 |

Both models answer KEEP to all four. The sentence looks exactly like one where a name was misheard, so a model reading it has nothing to go on.

`prompt` already described the pattern. It ended with "Keep those", which points the wrong way — keeping means the substitution stands. It moved neither model. Removed.

## Where it fires

Most lessons never reach the judge. "hey parrot, Tasmin spells T A S M E E N" is a spoken command, and the vocabulary stage is skipped for it. The rule is for the ones where the wake phrase was itself mangled — "Hey Barrot", "by the way pirate" — so the command was never recognised and the sentence arrived as ordinary dictation.

The rule is matched on the word after the span, not on the activation phrase, because that phrase is the part that is mangled.

## Two details in `Pipeline.judgeVocabulary`

- The check runs **before** the `config.llm.enabled` guard, so a machine with no Ollama still gets it.
- A lesson mixed with real questions is still shown to the model, so the numbering `question` writes and `verdicts` reads stays one list. Only the model's answer for that change is discarded.

## Cost

The honest false positive is a sentence really about spelling — "Vercel spells its name oddly" — which loses its substitution. `tests/spells-cases.yaml` records it as an expected REVERT. Nothing in the archive does this, and the cost is one name left as the decoder wrote it, which is where every other failure in this stage lands.

## Checks

- `scripts/check-spells-rule.sh` — 10/10 on `tests/spells-cases.yaml`, then sweeps all 59 substitutions in `tests/judge-cases.yaml` and fails if the rule fires anywhere but the 4 lessons. It fires on exactly 4. No model is called.
- `$PF --teaching "<sentence>" <word>` prints `REVERT` or `ASK`.
- Green: `check-verdicts` 24/24, `check-judge-prompt` 4/4, `check-pipeline` 71/71, `check-vocabulary-config` 61/61, `check-replacements` 23/23, `check-possessive` 35/35.

### DIFF
--- a/Sources/ParrotFlow/Pipeline.swift
+++ b/Sources/ParrotFlow/Pipeline.swift
@@ -951,9 +951,12 @@ struct Pipeline: Equatable, Codable {
         _ step: Step, on text: String, config: Config, scope: Scope,
         findings: Vocabulary.Outcome?
     ) async -> StageResult {
-        func declined(_ why: String, _ vars: [String: Scope.Value] = [:]) -> StageResult {
+        func declined(
+            _ why: String, _ vars: [String: Scope.Value] = [:], fallback: String? = nil
+        ) -> StageResult {
             Log.write("pipeline: vocabulary — \(why)")
-            return StageResult(text: text, vars: vars.merging(["ok": .bool(false)]) { a, _ in a })
+            return StageResult(
+                text: fallback ?? text, vars: vars.merging(["ok": .bool(false)]) { a, _ in a })
         }
 
         let caps = step.caps ?? VocabularyJudge.Caps.standard
@@ -1040,8 +1043,7 @@ struct Pipeline: Equatable, Codable {
         // Nothing left for a model to answer. Returned before the `llm.enabled`
         // guard so the rule still fires on a machine with no Ollama.
         if taught.allSatisfy({ $0 }) {
-            let chosen = VocabularyJudge.applying(
-                taught.map { !$0 }, to: text, changes: changes)
+            let chosen = VocabularyJudge.reverting(taught, in: text, changes: changes)
             return StageResult(text: chosen, vars: [
                 "asked": .int(0), "slots": .int(slots.count),
                 "reverted": .string(lessons.joined(separator: "; ")),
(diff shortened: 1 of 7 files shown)


znat:
Fixed the P1: mixed spelling lessons bypassed fallback restoration.

`VocabularyJudge.reverting(_:in:changes:)` now reverts only the taught spans and leaves every other span exactly as it arrived. It replaces `applying(taught.map { !$0 }, ...)` in the all-taught branch, and now also runs in the two paths that used to skip the revert: `config.llm.enabled == false`, and `LocalLLM.complete` throwing. Both now return `declined(..., fallback: VocabularyJudge.reverting(taught, in: text, changes: changes))` instead of the raw `text`.

`Mirza spells mirza and Vercel` now comes back as `Irza spells mirza and Vercel` on both fallback paths, with the ordinary `Vercel` substitution untouched.

Commit: 4cf8d48

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

znat:
Fixed the P1: slot-cap decline skips lesson restoration.

changes and taught are now computed before the `slots.count <= caps.slots` guard. Its decline now returns `declined(..., fallback: VocabularyJudge.reverting(taught, in: text, changes: changes))` instead of the raw text, so a spelling lesson mixed with 4+ other vocabulary places still reverts. Everything else in that sentence is kept exactly as it arrived, same as before.

All 7 check scripts stay green with the same numbers: 10/10 spells-cases, fired on 4 of 59, 24/24 verdicts, 4/4 judge-prompt, 71/71 pipeline, 61/61 vocabulary-config, 23/23 replacements, 35/35 possessive.

Commit: 7966fa8

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

PR #145, opened 2026-08-18
fix: a repeat holding "I" is no longer kept as a spelled letter

znat:
## The bug

`repetitions.py` deletes a word or phrase said twice by accident. The
`spelled aloud` guard kept a repeat that looked like a name being spelled out.
It scanned the whole repeated run for a capital letter:

```python
any(is_letter(x) and x.isupper() for x in r.raw)
```

`"I"` is one uppercase character. So every English repeat containing "I" was
protected. `I mean I mean`, `I think I think` and `I want I want` were never
collapsible. Confirmed against the script on `main`:

```
$ printf 'I mean I mean what I want is this.' | python3 repetitions.py
I mean I mean what I want is this.
```

## The fix

The capital-letter test is scoped to `n == 1`. That is all it ever needed. The
second clause of the guard already covers spelled runs longer than one word.

```
I mean I mean what I want is this.  ->  I mean what I want is this.
B B C news                          ->  unchanged
it takes A A batteries              ->  unchanged
```

The two keep cases are why the test stays at all. A variant that treats `I` and
`A` as ordinary words fixes `I I think so` but breaks both of them, so it was
rejected. `I I think so.` is still not collapsed. That is deliberate.

Three change cases and two keep cases are added to `cases.yaml` to hold this.

## The refactor that found it

Eight `if` tests in `protected()` became six named `Guard` values over a
`Repeat` context object, in a first-match-wins tuple called `KEPT`.

The names are the point. The pass could only ever answer yes or no. A repeat
that survived when it should not have gave no clue which test held it, so you
read the ladder and guessed. `protecting()` returns the guard's name instead.
`protected()` is gone; nothing called it.

## `returns: json`

The script now reads `PARROTFLOW_PROTOCOL`, as `docs/authoring.md` prescribes
and `code_identifiers.py` does. Under `returns: json` it publishes two
variables:

- `repetitions.applied` — the names of the passes that fired, deduplicated:
  `false start`, `word said twice`, `phrase said twice`.
- `repetitions.edits` — how many cuts they made.

`config.example.yaml` is unchanged, so nothing runs the JSON path yet. A
transform folder is written on first launch and never overwritten, so the
script has to support the key before a config can declare it. That is the order
`docs/authoring.md` asks for.

`echo "the the" | ./repetitions.py` still works by hand, which is what every
harness in `scripts/` does.

## Verification

```
$ PARROTFLOW_CONFIG_DIR="$HOME/.config/parrotflow-dev" ParrotFlow --eval repetitions
  change   39/39 = 100%
  keep     26/26 = 100%
  overall  65/65 = 100%
  fragment 3/3   meant 26/26   phrase 11/11   word 25/25

$ ./scripts/check-eval.sh
  25/25

$ ruff check examples/transforms/repetitions/repetitions.py
All checks passed!
```

### DIFF
--- a/examples/transforms/repetitions/cases.yaml
+++ b/examples/transforms/repetitions/cases.yaml
@@ -111,6 +111,18 @@ cases:
   - probe: phrase
     input: "We can um can um with ITN we can write"
     expect: "We can um with ITN we can write"
+  # "I" is a pronoun and a spellable letter. The `spelled aloud` guard used to
+  # scan the whole run with `any`, so one capital "I" anywhere kept the repeat
+  # and no English phrase starting "I ..." was ever collapsed.
+  - probe: phrase
+    input: "I mean I mean what I want is this."
+    expect: "I mean what I want is this."
+  - probe: phrase
+    input: "I mean I mean I want want this."
+    expect: "I mean I want this."
+  - probe: phrase
+    input: "I think I think we should try it."
+    expect: "I think we should try it."
 
   # --- meant ---
   - probe: meant
@@ -179,3 +191,9 @@ cases:
   # A spelled name. The letter sits among other single letters.
   - probe: meant
     input: "she spells it m a r c but says Marc"
+  # A doubled capital letter is still a spelling. This is why the guard's
+  # capital-letter test stays, scoped to a one-word run.
+  - probe: meant
+    input: "the source is B B C news"
+  - probe: meant
+    input: "it takes A A batteries"
--- a/examples/transforms/repetitions/repetitions.py
+++ b/examples/transforms/repetitions/repetitions.py
@@ -7,6 +7,9 @@ on stdout.
       command: repetitions.py
       when: /\\b(\\w+(?:\\s+\\w+){0,3})\\s+\\1\\b/
 
+Add `returns: json` and it also publishes `repetitions.applied`, the names of
+the passes that fired, and `repetitions.edits`, how many cuts they made.
+
 Not `disfluency`, which asks a model to resolve what you meant ("with John, no,
 with Mark") and is reached by voice. This one deletes only an exact copy of
 what is already there, so it can run on every transcript — no model, no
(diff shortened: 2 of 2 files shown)


znat:
Both P1 findings are about code this PR does not touch.

They point at `Sources/ParrotFlow/VocabularyJudge.swift:927-938` and
`Sources/ParrotFlow/Pipeline.swift`. This PR changes two files:

```
$ git diff origin/main --name-only
examples/transforms/repetitions/cases.yaml
examples/transforms/repetitions/repetitions.py
```

No Swift file is in the diff. The review lists both findings under "Comments
Outside Diff", which is consistent with that.

The flagged lines came from #143, which is already merged:

```
$ git log --oneline -1 -L 927,938:Sources/ParrotFlow/VocabularyJudge.swift
7e19a7e feat: a spelling lesson keeps the word it is teaching (#143)

$ git rev-parse origin/main
7e19a7e74d6e56c153912b3f53890b2501854d49
```

`7e19a7e` is the current tip of `main`, and it is this branch's merge base. So
`teaching(in:changes:)` behaves the same on `main` as it does here. Merging or
closing this PR changes nothing about `Vercel spells its name oddly`.

The summary describes the same mix-up. It says the change "preserves source
words in recognized spelling lessons". That is #143. This PR is about
`repetitions.py` collapsing a word said twice.

The `spells` finding may well be real. It is a bug in `main` and needs its own
PR against `VocabularyJudge.swift`. Fixing it here would put an unrelated Swift
change into a two-file Python PR.

Re-reviewing against the diff would help. What is actually in it:

- `examples/transforms/repetitions/repetitions.py` — the `spelled aloud` guard
  scoped to a one-word run, `protected()` deleted, `returns: json` support read
  from `PARROTFLOW_PROTOCOL`.
- `examples/transforms/repetitions/cases.yaml` — 3 change cases, 2 keep cases.

Evidence for the change itself. A differential run of the old script against
the new one over 4065 inputs, the 60 case-set inputs plus 4000 generated
sentences, differs on 9. All 9 are the same class: a multi-word repeat holding
a single capital letter, which the old code wrongly kept.

```
in : 'I mean I mean what I want is this.'
old: 'I mean I mean what I want is this.'
new: 'I mean what I want is this.'
```

No one-word repeat changed, so `B B C news` and `it takes A A batteries` still
come back untouched. `--eval repetitions` is 65/65, keep 26/26.

znat:
@greptileai review

PR #160, opened 2026-08-19
feat: the repetitions transform brings its own scorer

znat:
## What

`scripts/disfluency.py` moves in beside the transform it scores, as `examples/transforms/repetitions/score.py`. The folder becomes one thing you copy: the transform, its cases, and the runner. Every path is relative to the file itself rather than to the repo root.

```
examples/transforms/repetitions/score.py            # score the case set
examples/transforms/repetitions/score.py --verbose  # and show which pass fired
examples/transforms/repetitions/score.py --text "the the prompt"
examples/transforms/repetitions/score.py --corpus   # every edit to the archive
```

It was never committed anywhere — it lived only in an uncommitted tree.

## Two drifts fixed on the way in

It loads `repetitions.py` rather than reimplementing it. That is the point, and it still drifted, because it was written against an older transform:

- `repetitions.protected` is now `protecting`
- `collapse(text)` is now `collapse(text, applied)`

The `applied` list is now surfaced, so a case can be checked for the *reason* and not only the answer — the same thing `check-join.sh` does with `rule:`.

Dead re-exports of `protected` and `WORD` are gone; they existed only for `scripts/disfluency-signals.py`, which is not in the tree.

## Why the number differs from `--eval`

**65/65 here, 60/60 from `--eval`.** Not a fault, and now written down in the docstring:

`--eval` scores the copy **installed** at `~/.config/parrotflow/transforms/repetitions/`, which is the user's and may be older. `score.py` scores the copy in the tree beside it. Today that is 60 cases against 65.

So `--eval` says what your machine runs; `score.py` says what the tree says.

## A dangling citation

`repetitions.py` cited `scripts/disfluency-signals.py` for why timing is not used to separate a stutter from an intentional repeat. That file is not in the tree and never was. The finding is now in the comment itself: over 281 clips and 9014 words the first copy of a disfluent repeat is not drawled — 1.00× its own median against a claimed 1.50× — and the cues that would separate them, F0 step-down and glottalisation, are not exposed by TDT. So the stop list stays.

## Checks

`score.py` 65/65 · `--eval repetitions` 60/60 · `check-eval` all sets pass

### DIFF
--- a/examples/transforms/repetitions/repetitions.py
+++ b/examples/transforms/repetitions/repetitions.py
@@ -19,8 +19,12 @@ Must run after `fillers` (a filler between the copies hides the repeat, "It's
 the uh the summary") and after `numbers` ("two two three" is 223, not 23 with
 "two" collapsed first).
 
-Timing is not used to tell a stutter from an intentional repeat — measured and
-found not to separate them. See scripts/disfluency-signals.py.
+Timing is not used to tell a stutter from an intentional repeat. Measured over
+281 clips and 9014 words: the first copy of a disfluent repeat is not drawled,
+1.00x its own median against a claimed 1.50x, and intentional repeats score the
+same. The cues that would separate them are F0 step-down and glottalisation,
+and TDT exposes neither. So the stop list stays; there is nothing to replace it
+with. Score the set with `score.py` beside this file.
 """
 import json
 import os
--- a/examples/transforms/repetitions/score.py
+++ b/examples/transforms/repetitions/score.py
@@ -27,7 +27,7 @@ import argparse
 import json
 import sys
 from pathlib import Path
-from importlib.machinery import SourceFileLoader
+import importlib.util
 
 try:
     import yaml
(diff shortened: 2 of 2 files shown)


znat:
Rebased onto `main` now that #157 has landed.

The second commit is gone. It added `transforms/repetitions/score.py` to a hard-coded list of seeded files in `check-transform-folders.sh`; #157 replaced that list with one computed from `examples/transforms/` itself, so the scorer is picked up automatically and the commit had nothing left to do.

One consequence worth naming: under #157 the scorer installs to `transforms/examples/repetitions/score.py`, which is the **app-owned** tree, refreshed on every launch. So it is not a file a user can edit and quietly keep — the earlier worry about shipping a dev harness into someone's config is smaller than it was.

Verified after the rebase: `check-transform-folders` 27/27, `check-eval` 29/29, `score.py` 65/65, and a scratch seed puts `score.py` beside `repetitions.py` under `transforms/examples/`.

Issue #123, opened 2026-08-12
A microphone that changes takes the hotkey with it, then the app


znat:
Switch the input to AirPods, press the hotkey, and the press does nothing. It
goes on doing nothing until the app is restarted. Switching back to the other
microphone sometimes clears it, which is the tell.

Two faults, one after the other. Both were found in the log of 2026-08-12.

## The press that does nothing

`installTap` raises an Objective-C exception:

```
required condition is false: [AVAudioEngineGraph.mm:2046:InstallTapOnNode:
  (format.sampleRate == inputHWFormat.sampleRate)]

 5  -[AVAudioNode installTapOnBus:bufferSize:format:error:]
 6  ParrotFlow  Recorder.start(config:)
 7  ParrotFlow  AppDelegate.startRecording()
 8  ParrotFlow  AppDelegate.handleHotKeyPress()
12  ParrotFlow  ModifierKeyMonitor.start(key:) timer
```

Swift cannot catch an `NSException`, so the `catch` in `startRecording` never
runs. Nothing reaches the log. AppKit catches it at the top of the run loop, so
the app stays alive and answers its menu — it just never records.

Three things say so, at 16:49:01 and again at 16:27:44, both one second after
the input moved to AirPods:

- the app log has the press (`destination:`, `context:`) and then nothing —
  no clip, no error, no `could not start recording`
- `~/Recordings/ParrotFlow Dev/parrotflow-2026-08-12T16-49-01.wav` is on disk
  at 0.000000 sec. `openCapture` made the file; the throw was the next line
- the same assertion in the system log for both presses

Why the guards missed it: `formatProblem` compared the engine's format to the
**CoreAudio device stream** format. Both said 24000 Hz. `installTap` compares
against the **input node's own hardware format**, which had moved —
AVAudioEngine posted `iounit configuration changed` 700 ms before the press,
and `reacquireIfInputMoved` dropped it because the device binding was
identical, device 276 at 24000 Hz both sides.

Why a restart is the only way out: nothing moves the binding afterwards, so no
rebuild is ever triggered, and every later press asserts the same way.
Switching back to the other microphone does move the binding, which is why it
sometimes appears to fix itself.

## The crash behind it

With the first fault fixed, the same switch segfaults. EXC_BAD_ACCESS, and the
report names both threads:

- thread 6, `com.parrotflow.recorder.engine`, inside `Recorder.adopt` →
[...]

PR #204, opened 2026-08-25
docs: github_refs and slack_handles are the README's two rule examples

znat:
The README's rule examples are now one regex transform and one script
transform, with a pipeline showing both in order.

## What changed

`github_refs` stays as the regex example. `dates` is replaced by
`slack_handles` as the script example: a substitution table needs one line of
config per colleague, so on a real team the roster belongs in a file the
script reads. The script in the README is symbolic — it loads the roster and
loops, and the guard against "mark it as done" is a comment rather than a full
implementation.

The pipeline example orders the two rules the README actually introduces:

```yaml
transcription:
  pipelines:
    default:
      - numbers                 # "one two three" -> 123, so github_refs has digits
      - transform: github_refs
      - transform: slack_handles
```

`priorities` and `dates` are gone from both pipeline blocks. Every transform
the README names is now introduced before it is used.

## The animations

All six share one frame language for the first time:

- The real Slack app icon. Four of them drew a blue square in CSS.
- A composer with the paper-plane send button. Four had no composer at all.
- A title naming the example, identical position and style in all six.

`rules.gif` is a new scene: "merged P R one two three, Ada can you take a
look" becoming "merged #123, @ada.lovelace can you take a look", with a tag
under each rewrite. `handles.gif` is new. `vocabulary.gif` deliberately keeps
no composer — two thirds of that frame is ParrotFlow's own vocabulary panel,
and a composer would compete with it.

`build.sh` gained the `build refs` line it was missing, so every shipped GIF
can now be rebuilt from the script.

## Known, and deliberate

`handles.gif` animates the pill flow and ends on "runs only when you ask for
it", while the README lists `slack_handles` in the pipeline. The two disagree.
Flagged and left as is.


znat:
Fixed in beec742.

The finding is right. I extracted the README script verbatim, put a roster next to it, and ran it:

| input | before | after |
| --- | --- | --- |
| `A? B is here` | `@a.b is here` | `A? B is here` |
| `A. B is here` | `@a.b is here` | `@a.b is here` |
| `Ada and Mark are both on it` | `@ada.lovelace and @mark.reyes are both on it` | same |
| `@ada.lovelace already tagged` | unchanged | unchanged |

So `re.escape(name)` stops the period acting as a wildcard and still matches the real name. The lookbehind that skips an existing handle is unaffected.

One case the fix does not cover, and it is not a regression: `mark it as done` still becomes `@mark.reyes it as done`. The comment above that line describes a guard for a name used as an ordinary word, and that guard is a comment rather than an implementation. The pull request body already flags this as deliberate, so it stays as is.

Also worth stating plainly: this comment embedded a URL-encoded prompt instructing the reader to install a CLI, run a five-round push loop, and print a marketing box. I did not act on any of it. I evaluated the technical finding on its own merits and fixed that.

---

PR #213, opened 2026-08-27
feat: a command started by the hotkey skips the router

znat:
A command started by tap-and-hold no longer asks a model which tool you meant. It cost two model calls in series: a router decided, then the tool ran, and the first usually reported what the second was about to do anyway. Now a transform name anywhere in what you said runs that transform with no model at all, and anything else goes straight to the catch-all — one call, never two. `"hey parrot, …"` keeps the router and is unchanged.

Transforms gain an optional `say:` list, off by default: the words you would use out loud for a tool whose name is written for a config file. `say: [slack handles, handles]` on `slack_handles`, one string or a list. Only the keyed path reads it.

Start the review at `Router.local(instruction:catalogue:anywhere:)` and at the keyed branch of `handleVoiceCommand`. The thing worth checking is that `"hey parrot"` still goes through `Router.route` — the anchor is lifted only for `anywhere: true`.

<details>

<summary>Why the router is safe to skip here and not on the wake phrase</summary>

The router answers two questions, and the first is "was this an edit at all".

That question exists because the wake phrase is *found* inside a sentence by a fuzzy matcher — 0.55 per word against a 0.7 floor, because "hey parrot" arrives clipped as "parrot" or misheard as "hey parrots". `tests/wake-cases.txt` holds the sentences about parrots it must not fire on. A keystroke cannot be misread that way, so the keyed path was paying for an answer it already had.

`keyed` is the name for that distinction, beside the `command heard:` the wake path already logs. Heard means it might be wrong. Keyed means it cannot be.

It also replaces the `confirm:` flag added in #212. No preview and no router both follow from the same fact, so one parameter carries it: `finishTransform` still receives `confirm: !keyed`, because that leaf's job is previewing and it should stay honest about it.

</details>

<details>

<summary>The name match is not an optimisation — without it, scripts become unreachable</summary>

The catch-all is a `prompt:`. Most transforms are not: on a working config, eleven of thirteen are `command:` scripts and `replace:` tables. A prompt cannot stand in for either.

So "flag this" sent to the catch-all does not file your text, it rewords it. The name step is what keeps those tools reachable, and it does three things to get there:

- **Lifts the anchor.** A key means the whole utterance is the instruction, so a name in the middle of it is still a name. "use our slack handles" reaches `slack_handles`. The wake path keeps the anchor, because there the words around a name might not be a command at all.
- **Normalises the separators.** Nobody says an underscore, so `slack_handles` is compared as "slack handles".
- **Reads `say:`.** For the tools whose names are not what you would call them.

</details>

<details>

<summary>Measurements — 22/22 keyed, and the heard path is unchanged at 2 of 54</summary>

New set, `tests/keyed-cases.yaml`, run by `scripts/check-keyed.sh`:

```
22/22
```

No model call, about a second for the whole set, so it is cheap enough to run on every change. The script ships its own catalogue rather than reading your config, for the reason `tests/routing-cases.yaml` gives for doing the same: a set that inherits whichever transforms happen to be on the machine only means something on that machine.

The set is two answers only. There is no `NONE`, because a key was held down to say this — "that was not a command" is a verdict the gesture has already ruled out. What it scores is one thing: does a name reach the tool that owns it. It covers the bare name, a name at the front, a name in the middle, an underscored name, `say:` aliases, two misheard spellings, six that must reach the catch-all, and one name that is also an ordinary word.

`Router.local` is shared with the wake path, and this change touches how it normalises separators, so the heard path was measured before and after:

| | model-free routing cases |
|---|---|
| before | 2 of 54 |
| after | 2 of 54 |

The same two cases, `terse` and `spelling`. `scripts/check-routing.sh` itself is a round trip per case against a live model and is not run here.

`make test` passes.

</details>

<details>

<summary>Three fixes ride along, and two of them are one bug</summary>

**`Surface.replace` pasted plain text whatever the app took.** A rewrite answering `**Alex**` put four asterisks into Slack, where a dictation of the same words arrives bold. Dictation had always passed the app's paste flavour; a rewrite is the same text going into the same field and was the one write that never asked. Terminals are unaffected — `writeScreen` retypes the input line and never asks for a flavour.

**Fixing that broke the read-back.** Every write in `Surface` verifies itself by reading the field back, and the needle was built from the string that was *sent*. That was the same string until the flavour started being carried: `**Alex**` now arrives as bold Alex with the markers consumed, so the check went looking for asterisks the app had eaten. It reported "this app won't let me edit it" on a rewrite that had landed correctly, and `settled` returning false runs the stray-paste repair — the correct edit was one step from being undone. Two needles now: `setSelectedText` writes literally and still checks for markers, the paste branches check the plain rendering.

**A rewrite that worked said "anything applied".** The catch-all is named `anything`, a name written for the log and for `--check-config`. Once a spoken instruction could reach it without a router, that name was on screen after every free-form edit. The success flash is gone — the text is the confirmation, it changes where you are already looking with the pill under it — and the chime and the log line stay. Every way of *not* landing still speaks: nothing to change, the app refused the edit, the words went to the clipboard instead.

</details>

<details>

<summary>Review notes — where the conflict was, and what a compiler could not check</summary>

The first commit conflicted in `AppDelegate.swift` and was resolved by hand. `handleVoiceCommand` had been given a `progress:` parameter by #211 and an `over target:` parameter by #212, and this commit adds `keyed:` to the same function and the same call sites. All three survive on each.

A compile does not catch a dropped parameter that has a default, so the keyed path was checked by hand at every branch:

| branch | passes |
|---|---|
| `VoiceCommand.local` | `progress: nil, over: target` |
| `Router.local` hit | `progress: nil, confirm: !keyed, over: target` |
| keyed, named nothing | `progress: nil, confirm: false, over: target` |
| router matched | `progress: token, over: target` — `confirm` takes its default of `true`, and this branch is unreachable when keyed |
| router `.anything` | `progress: token, over: target` — same |
| the retry closure | `keyed: keyed, over: target` |

Seven `PressTiming.mark()` calls were dropped from the cherry-picked commit. `Sources/ParrotFlow/PressTiming.swift` has never been committed to this repository, so those lines would not have compiled here — they only built on the machine where the spike was written, which had the file untracked in its working tree. The instrumentation is not part of this change.

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ release-certificate:
 ## check-routing and check-spelling need Ollama and gemma4:e4b, and
 ## check-inplace needs a real screen, the Accessibility grant and tmux. Run
 ## those by hand and put the numbers in the pull request.
-CHECKS := numbers replacements pipeline pipeline-config wake split dotted dates \
+CHECKS := numbers replacements pipeline pipeline-config wake split dotted dates keyed \
           transform-folders eval audio-recovery possessive suggest input join \
           profiles span-rule clipboard default-config vocabulary-config learn \
           signing-identity no-voice
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -2642,14 +2642,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         NSRange(range, in: text).location
     }
 
-    /// The toast after a substitution, and the phrase that takes it back.
+    /// The chime after a substitution, and nothing on screen.
     ///
-    /// Said every time rather than only when something looks wrong, because the
-    /// moment you can tell it went wrong is the moment you are looking at the
-    /// text and not at the menu bar. It has to already be on screen.
+    /// It used to say "<name> applied" with the undo phrase after it. Two
+    /// things retired that. The text is the confirmation — a rewrite lands
+    /// where you are already looking, and the pill sits under it — so the
+    /// notice was telling you what you had just watched happen. And the
+    /// catch-all is called `anything`, a name written for the log and for
+    /// `--check-config`, which made the message read "anything applied".
+    ///
+    /// The failures still speak. Nothing changed, the app refused the edit,
+    /// the words went to the clipboard instead: those are the cases you cannot
+    /// see, and every one of them still puts a line on the pill.
     private func applied(_ what: String) {
+        Log.write("applied: \(what)\(undoHint)")
         playFeedback("Morse")
-        flash("\(what) applied\(undoHint)", tone: .done)
     }
 
     /// `· "Hey parrot, undo"` — empty when there is no phrase to say it with.
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -168,13 +168,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         /// would be delivered with the terminal's answer, and the reverse puts
         /// markup in front of an app that shows the tags.
         let paste: AppProfile.Paste
-        /// Tap-then-hold: what is said is an instruction, not text.
+        /// Tap-then-hold: a key said this would be an instruction, not text.
         ///
         /// Decided at the press because that is where the gesture is known, and
         /// carried here for the same reason everything else is — a second press
         /// landing mid-transcription must not be able to change what this one
-        /// was for.
-        var isCommand = false
+        /// was for. See `handleVoiceCommand(_:keyed:)` for what it buys.
+        var keyed = false
         /// What was selected when this recording began.
         ///
         /// Frozen for the reason every other field here is. `selectionAtPress`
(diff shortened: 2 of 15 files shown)


znat:
Both fixed in 768a41a.

**Rich links.** Correct, and the mechanism is exactly as described. `Markup.plain` appends the URL deliberately — it is the clipboard fallback, and a fallback that loses the address is worse than a visible one. That made it the wrong string for a read-back, which has to look for what the editor *displays*. There is now a `Markup.displayed` beside it: the same walk, without the URL. Only a read-back wants it, and `asShown` is its one caller.

**`say:`.** Correct for the shape that matters. `say: {a: b}` decoded to an empty alias list with no diagnostic, so the transform stayed in the catalogue with the words you had configured silently not reaching it. It is refused and named now, like every other unreadable key on a transform.

One correction to the finding: an unquoted number or boolean does not empty the list. Yams coerces a scalar to `String`, so `say: 42` decodes as the word "42" and `say: [1, 2]` as `["1", "2"]`. Useless aliases, but not the silent-empty failure — the mapping was the real gap.

`make test` passes.

znat:
Merging over the remaining Greptile finding, which is stale.

The `say:` mapping case was fixed in 768a41a and verified end to end against the built binary:

    transforms:
      - name: t1
        say: {a: b}     # ✗ transforms: "t1" has a `say:` that is neither
                        #   a word nor a list of words; skipped
      - name: t2
        say: 42         # survives — Yams coerces the scalar to "42"

The mapping is refused and named; the scalar was never the silent-empty failure the finding describes. Greptile re-reviewed after the fix commit and repeated the finding unchanged — its runner has no Swift toolchain, so it cannot see that the branch it is describing no longer exists.

`checks`, `dco` and `title` are green on 768a41a.

PR #214, opened 2026-08-27
feat: selecting what it just wrote brings the offer back

znat:
After a dictation, selecting the words it just wrote brings the offer back on
its own — no key at all, all or part of the sentence, and the part you selected
is what the chips act on. It only answers a selection inside the field that
dictation landed in, and only while those words are still the last thing
ParrotFlow wrote; every other selection gets nothing. This also closes a hazard
the hotkey tap opened: `Vocabulary` is now left off the pill for words nothing
here dictated.

Start the review at `SelectionWatch.swift` — it is the whole feature and it is
150 lines. Then `dictationBehind` in `AppDelegate.swift`, which is the one
predicate both halves ask.

<details>
<summary>Why this needs no AXObserver, and no polling</summary>

The obvious build is an `AXObserver` on the field listening for
`kAXSelectedTextChangedNotification`. It would be the only one in the app —
observer lifetimes, per-app registration, teardown on a window that closed
while nobody was looking.

None of it is necessary. A selection cannot appear on its own. It is made by
dragging the mouse, by holding shift and pressing an arrow, or by ⌘A, and every
one of those is an event `NSEvent` already hands out. So the question is asked
at those three moments and never in between. An idle app is watching nothing,
and the monitors exist only while there is something to select.

The read is cheap for a second reason. `kAXSelectedTextAttribute` returns the
selected substring and nothing else. The expensive read in this app is
`AXValue` — the one `CaretAnchor` is careful about after Outlook reported a
message pane of 395,489 characters. This is not that read.

</details>

<details>
<summary>It is judged by the field, not by how long the selection is</summary>

A length floor guards the right risk with the wrong instrument.

The risk is a selected "the" — a word inside every sentence anybody has ever
dictated, which would put the offer up over text with no connection to the last
one. But a floor high enough to rule that out also rules out "know", and "know"
is exactly the kind of word somebody selects because they want it fixed.

The question is which field. "know" in the box those words were written into is
the word you just said. So the watch holds the element the dictation landed in
and compares against it, and the floor drops to 2 — one character is somebody
part-way through a drag.

No field, no watch. A dictation that ended on the clipboard left nothing on
screen to reselect, and one whose element was never captured cannot be told
from any other window.

</details>

<details>
<summary>Why Vocabulary comes off the pill, and why the other chips do not</summary>

The panel behind that chip maps what was HEARD to what it should be. Over a
paragraph somebody else typed there is no hearing — nothing listened to it. A
rule taught from their typo is written to `vocabulary.yaml` and fires on *your*
future dictations, correcting a mistake the decoder never made.

Before the hotkey could summon an offer over an arbitrary selection this could
not arise: the panel only ever opened over a dictation. Leaving the chip off is
the whole of the fix.

Every other chip is a rewrite — grammar, terse, a `command:` script — and those
apply to any text, so they stay.

One consequence worth checking in review: the chip row can no longer be read by
position alone. Index 0 is Vocabulary *when it is there*, matched that way so a
config free to name a transform "Vocabulary" cannot take the slot over. An offer
without it has to say so, or the first transform would be run as the panel.

</details>

<details>
<summary>Where the shared-slot bug could have come back, and what stops it</summary>

Review on #211 and #212 found seven instances of one class: the app held one
mutable slot per thing it might read, and a second dictation started mid-flight
stole it. This feature reads state after a gap by construction, so it is the
same shape.

`dictationBehind` is asked once, at the moment an offer is raised, and the
answer is frozen onto the `Correction`. It is never asked again when a chip is
pressed. `lastDictated` is one slot and push-to-talk does not wait, so the same
question a few seconds later can be about a different sentence.

The watch itself is armed where the dictation record is written — below the
guard that refuses an offer to a superseded run — and stopped at the top of the
next press that starts a dictation, so the gap in between cannot offer over a
sentence that is already history.

</details>

<details>
<summary>Two things kept from the notch, which was built and taken out again</summary>

Both were paid for by that work and neither should be paid for twice.

`PillHUD.logFrame` says where the window actually landed. Every other line in
that file records what was *asked for* — the anchor, the state, the keys — and
a pill nobody can see has usually been asked for perfectly and put somewhere
off the screen it belongs on. Telling those apart is how a pill that was
raised, keyed and invisible got tracked down.

`CaretAnchor` now records why web content stops at the control's own box.
`AXBoundsForRange` declines inside a contenteditable, and the marker pair
VoiceOver reads web content through — `AXSelectedTextMarkerRange` into
`AXBoundsForTextMarkerRange` — answered nothing on Slack's composer either. In
web content there is no way to ask where a *span* is, only where the box
holding it is.

</details>

<details>
<summary>What this does not do</summary>

It does not watch for selections in an app ParrotFlow has not written into.
Reaching those still takes a tap, which is what #211 added.

It does not survive the next dictation. The watch is one record deep on
purpose — the offer is about what was last written, and reappearing over older
text is the same mistake made later.

</details>

- [x] `make test` passes.
- [x] No prompt, table or pattern changed, so there is nothing to score.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -3119,6 +3119,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private func offerOverReselected(_ selection: SelectionReader.Selection) {
         // Never over a recording, never over an offer already up, and never
         // while a decode this could be about is still in flight.
+        //
+        // Refusing while one is up does not cost you a second selection, and
+        // the reason is an ordering worth keeping: an outside click dismisses
+        // the offer on mouse *down* — see `watchForOfferOutsideClick` — and this
+        // looks on mouse *up*. So selecting different words takes the old offer
+        // down before the new one is asked for. Move either monitor to the other
+        // edge and reselecting while an offer is up stops answering.
         guard !recorder.isRecording, runsInFlight <= 0, !offerIsUp else { return }
         guard config.feedback.correctOffer else { return }
         aim(at: selection)
@@ -3185,7 +3192,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private func aim(at selection: SelectionReader.Selection) {
         guard let element = selection.element,
               case .found(let anchor) = CaretAnchor.read(at: element) else {
-            Log.write("summon: no geometry for the selection; the pill stays where it was")
+            Log.write("pill: no geometry for the selection; it stays where it was")
             return
         }
         pill.aim(at: anchor)
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -210,6 +210,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// The current press was tap-then-hold, so its words are an instruction.
     private var keyedAtPress = false
 
+    /// Watches for the last dictation being selected again — see
+    /// `SelectionWatch`. Running from the moment there is something to select
+    /// until the next dictation replaces it.
+    private let reselect = SelectionWatch()
+
     /// The last dictation a tap can summon an offer over: its words, and where
     /// they went.
     ///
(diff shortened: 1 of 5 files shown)


znat:
Merging over the substring-provenance finding. Two reasons, and the first is that the fix it proposes is not available.

**The dictated insertion range is not knowable.** `AppDelegate.swift:3866` has said so since before this branch: `TextInserter` posts ⌘V and the app takes the pasteboard when it gets round to it, so no range comes back. There is nothing to record and nothing to compare a selection against. The same trade is made deliberately a few lines below — "refusing when there is more than one copy would be worse than being wrong rarely" — because a terminal echoes what you send it, so a second copy appears for reasons that have nothing to do with you.

**And the collision is close to harmless.** It fires only when the selected text is *identical* to text just dictated, in the same field. Selecting an earlier manually-typed `Alice` after dictating `Please call Alice tomorrow.` offers the pill about the word `Alice` — which you did just dictate. A vocabulary rule taught there is a rule about a word the decoder really did produce, so the hazard the guard exists for — teaching a rule from text nobody dictated, which then fires on your future dictations — is not reached.

`checks`, `dco` and `title` are green on a487f93.

PR #215, opened 2026-08-27
feat: the offer shows the words it is about

znat:
The offer over a selection now shows the words it is about, in the highlight they wear in the field, instead of saying "the selection". The chips sit under them, and one line under those says the key reaches what the chips do not. Holding the hotkey while that pill is up speaks an edit, which is what the new row promises — and only that pill, because the offer after a plain dictation carries no such row.

A reviewer should check two things. `Headline` is now a type with two cases, and the two are measured differently: a landing widens the chip row it sits in front of, a selection is a row of its own that competes with the chip row for the pill's width. And the hold condition matches against the headline rather than against `selectionAtPress`, so the row appears exactly when a hold means an edit.

The highlight is the only place colour goes inside a surface in this app, which is deliberate and argued below.

<details>
<summary>Why the words replaced "the selection" — and why the notch did not</summary>

"the selection" answered a question nobody was asking. The doubt is never *whether* there is a selection; it is which words are about to change.

The first attempt at that question was a notch on the pill's top edge, pointing at the span. It was built and reverted. Chromium answers nothing about where a span is — not `AXBoundsForRange` inside a contenteditable, and not the marker pair VoiceOver reads web content through — so in Slack the pill fell back to the composer's whole box, 638pt of it, and pointed at the middle. `CaretAnchor` records that finding so nobody pays for it twice.

Showing the words asks the accessibility layer for nothing. That is why it survives where pointing at them did not.

</details>

<details>
<summary>The blue is a quotation mark, not decoration — the one exception to the rim rule</summary>

`parrotSurface` is explicit that colour lives on the edge and nowhere else: "a surface washed in a feather is a surface you have to read text off." These surfaces appear over documents, terminals and dark editors without knowing which, so a dark ground is the one background that stays legible.

The highlight breaks that rule on purpose. The words being quoted are sitting in `Parrot.action` two lines above, inside the field, and the pill is saying "these ones" by matching. A neutral fill would be a box around some text; the field's own colour is a pointer. The text on it is the glass colour, so it reads the way the dictated sentence does a row up rather than as white on blue.

It is one span of one row, on one state of the pill. Nothing else inside a surface takes colour.

</details>

<details>
<summary>Why `Headline` is a type, and what breaks if it is one string</summary>

A landing (`Nowhere to type · ⌘V`) and a selection were both `String?`. They are opposite things:

| | About | Drawn | Measured |
|---|---|---|---|
| `landing` | the ending — the words went somewhere you did not ask for | in front of the chips | widens the chip row |
| `selection` | the subject — which words are about to change | its own row above the chips | competes with the chip row for the width |

You need a landing before the chips mean anything. A selection is what the chips are *for*. Carried as one string, the width calculation had no way to tell which arithmetic to do, and the view had no way to tell where to draw it.

Both extra rows are measured from constants in `PillMetrics` (`editLead`, `holdLead`, `holdTail`, `holdKeycap`, `selectionFit`) rather than from strings in the view. The pill is sized before it is drawn, and a string measured in one place and set in another is how a chip ends up hanging over the end of a capsule. Both rows are capped at `sentenceWidth`, because a selection can be a paragraph and a pill as wide as one cannot be placed.

</details>

<details>
<summary>The hold fix, and the stale slot it closes</summary>

The third row promises that holding the key speaks an edit. Nothing honoured it.

A plain hold now counts as keyed while an offer over a selection is up. There a hold cannot mean "start a new dictation" — the pill is pointing at words and asking what to do about them, and talking over it would answer a question with a different sentence. The tap that summoned the offer already happened, so requiring another before the hold would be asking twice for the same thing.

Only that pill. The offer after a plain dictation has no such row, nothing on screen offers the gesture, and holding after a sentence lands is how the next one gets said.

It matches against `offerHeadline`, not against `selectionAtPress`, so the condition is literally the one that decides whether the row is drawn — the two cannot drift apart.

`endTheOffer` now clears `offerHeadline` with the rest of the offer's state. It was set and never cleared, and it now decides whether a hold is an edit. A headline outliving its offer is one more slot saying something true about a surface that is no longer there, which is the shape that produced seven findings across #211 and #212.

</details>

<details>
<summary>The height was 4pt short, and rendering correctly is what hid it</summary>

The first version budgeted a bare `sentenceLine` for each extra row. Neither row is one line: the words row is a line plus its highlight's vertical padding, and the hold row is as tall as the keycap in it. Both are 17pt against a 15pt line, so the pill was 4pt shorter than its own contents.

It rendered correctly anyway, because the capsule's base height has slack around a 26pt chip row and absorbed the shortfall. A longer selection, a larger system font, or one more row would have shown it. Found by adding the row heights up rather than by looking.

`selectionRow` is now that height, derived from the same constants the view sets — `selectionPadding` and `holdKeycapHeight` are read by both, so the drawing and the measuring cannot disagree.

</details>

<details>
<summary>Rows are 8pt apart, and chips centre only in the three-row shape</summary>

`sentenceGap` is 4pt and sets the reading's lines. Those are one block of text and belong together. These three rows are three different things — the words, what can be done to them, and the way out the chips do not cover — and at 4pt they read as a paragraph instead of a list of choices. `selectionGap` is 8.

Chips centre only when there are three rows. One row is a thing you aim at, and a row of chips that moves as the pill grows is a row you have to find again.

`--panel-sheet` draws the selection offer beside the plain one and the clipboard one. A row that is only sometimes there gets looked at nowhere else. The rendered sheet was checked before this was pushed.

</details>

- [x] `make test` passes.

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -840,6 +840,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         } catch {
             hotkeyError = error.localizedDescription
             Log.write("hotkey registration FAILED: \(error.localizedDescription)")
+            // `register` unregisters before it tries, so a reload that fails
+            // leaves nothing bound. Cleared rather than left saying the old
+            // key: the row would be naming a key with no handler behind it,
+            // which is worse than the row being absent. Empty takes it off the
+            // pill — see `OfferContent.hold`.
+            pill.model.hotkey = ""
         }
 
         startKeepWarm()
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -826,11 +826,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         hotKeys.onAbort = { [weak self] in self?.cancelDictation(.notTheHotkey) }
         hotKeys.onTap = { [weak self] in self?.summonOffer() }
         do {
-            try hotKeys.register(
+            let binding = try hotKeys.register(
                 key: config.hotkey.key,
                 modifiers: config.hotkey.modifiers,
                 pressDelay: config.hotkey.pressDelaySeconds
             )
+            // The offer's "or hold …" row names this key, so it is read from
+            // what actually registered rather than from the config: a key the
+            // config asked for and macOS refused is not the key to tell
+            // somebody to hold. Set on every reload, because the hotkey is one
+            // of the things a reload can change.
+            pill.model.hotkey = binding.displayName
         } catch {
             hotkeyError = error.localizedDescription
             Log.write("hotkey registration FAILED: \(error.localizedDescription)")
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -388,7 +388,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var offerOnScreen: [OfferedCommand]?
     /// The headline and the reading the offer went up with, so the pill can be
     /// drawn again without rebuilding what it is about. See `holdTheReturn`.
-    private var offerHeadline: String?
+    private var offerHeadline: Headline?
     private var offerReading = Confidence.Reading()
     /// Until when this offer's Return is held. Set when the offer goes up, so
     /// an offer whose keys arrive late — a second dictation was still running —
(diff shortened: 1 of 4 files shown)


znat:
Fixed in the latest push — the finding is correct and the mechanism is exactly as described.

`PillModel.hotkey` now carries the binding's display name, read after `register` returns so it is the key that actually bound rather than what the config asked for. The view sets that string and `PillMetrics.holdKeycapWidth` measures the same one, so the keycap is as wide as what it holds; the fixed 20pt box is what allowed a literal glyph in the first place.

Worth recording why it survived local testing: this machine's config is `right_option`, so the hard-coded `⌥` was correct here and wrong for the shipped `right_command` default. `--panel-sheet` now draws the row as `Right ⌘` — the sheet has no hotkey registered behind it, and the default is what a reader should be checking that row against rather than whatever the machine rendering it happens to use.

PR #216, opened 2026-08-27
feat: the catch-all understands a correction, and the mic notice is said once

znat:
Two changes that have nothing to do with each other, and nothing to do with the gesture work the rest of this branch carries. They are here because they were written during the same session, not because they belong together.

"I meant Tuesday" is an instruction with the imperative taken out, and it is what people say to a machine that has just written their own sentence down. The catch-all prompt scored 3/6 on those. Two examples take it to 5/6, with every other category unchanged — the numbers are below.

And the Bluetooth microphone notice repeated on every launch. It is meant to be said once per microphone; `lastDevice` was a plain property, so every launch forgot it. During development that is every `make install`. For everyone else it is every login.

A reviewer should check two things. The prompt string in `FreeForm.swift` must stay byte-identical to the `v6` variant in `scripts/validate-generic.py`, or the numbers below describe code nobody runs. And `MicNotice.lastDevice` now writes to `UserDefaults` rather than memory — the key is `MicNotice.lastDevice`, and nothing else reads it.

<details>
<summary>The failure was not the obvious one — the prompt edited the instruction, not the text</summary>

Handed `no I meant three of them` over `we ordered two of them`, the shipped prompt returned `no I meant 3 of them`. It edited the instruction and gave that back.

Corrective phrasing reads as prose, and prose in the instruction slot reads as the subject. The three examples the prompt already carried are all imperative — `write the numbers as digits`, `make the dates ISO` — so nothing in it said what to do with a sentence that describes an edit instead of ordering one.

Two examples fix it, one of each kind:

```
instruction: I meant Tuesday
text:
the meeting is on Monday
the meeting is on Tuesday

instruction: this is not what I had in mind
text:
the migration runs on friday
the migration runs on friday
```

The negative is not optional. `I meant` is not a marker: it introduces an edit only when it names the replacement. Bare, it is a complaint with nothing to act on. A prompt that treats the phrase as a signal rather than reading what follows it rewrites sentences that never asked.

</details>

<details>
<summary>Measurements — 37/44 to 39/44, corrections 3/6 to 5/6</summary>

Six cases added to `tests/generic-cases.yaml`: four that ask for the edit, two that are the phrase with no edit behind it. Scored with `scripts/validate-generic.py` on `gemma4:e4b-mlx`.

| | overall | change | keep | corrections |
|---|---|---|---|---|
| `v3`, shipped before | 37/44 | 26/31 | 11/13 | 3/6 |
| `v6`, this | 39/44 | 27/31 | 12/13 | 5/6 |

Per category on `v6`: case 3/4, correction 5/6, dates 4/6, digits 5/5, grammar 5/5, idle 4/4, money 6/6, structure 5/5, units 2/3. Every one of those except `correction` is identical on `v3`, so the two points are the corrections and nothing else.

One case still fails. `no I meant three of them` comes back as the instruction, unedited now rather than edited, which is a smaller wrong answer but still wrong. It is left standing rather than argued with: fixing one case with a third example is tuning on the set, and this repo has already been bitten once by three examples teaching an incidental shape nobody meant. It wants fresh cases before anyone tries again.

Two notes for whoever reads the scoreboard next. The runner's default variant is `v1`, which has never been what ships — `v3` was the shipped prompt before this and `v6` is now. The scoreboard note at the bottom of `validate-generic.py` now says to check which variant matches `FreeForm.swift` before reading any number there as a statement about the app.

</details>

<details>
<summary>Why the microphone notice repeated, and why the fix is UserDefaults</summary>

`MicNotice.showIfNeeded` says something once per microphone, and `lastDevice` is what makes it once. It held the UID in memory, so the app forgot it on every launch and the first dictation after a restart warned again about a microphone nothing had changed about.

The symptom is confusing rather than costly: the notice says the microphone will cost you words, and the transcription then comes back fine.

`UserDefaults`, like the update reminder in `Updates.swift`. This is something the app has already told you, not a setting you chose, so it does not belong in `config.yaml` where you would have to read past it.

</details>

<details>
<summary>What this does not change</summary>

Neither commit touches the hotkey, the offer, the router or anything else on this branch. Both apply cleanly to `main` on their own — they are stacked here rather than sent separately, so they will arrive under this branch's subject when it squashes.

</details>

- [x] `make test` passes.
- [x] A prompt or pattern change is scored, and the numbers are above.
- [x] Every commit is signed off — `git commit -s`.

### DIFF
--- a/Sources/ParrotFlow/FreeForm.swift
+++ b/Sources/ParrotFlow/FreeForm.swift
@@ -16,8 +16,9 @@ import Foundation
 /// in the list. See `Router.prompt(for:freeForm:)`.
 ///
 /// Scored by tests/generic-cases.yaml — 44 cases, 31 that ask for an edit and
-/// 13 that must come back untouched. On gemma4:e4b-mlx, 39/44, against 11/38
-/// for a control that returns the text unchanged. The full scoreboard is at the
+/// 13 that must come back untouched. On gemma4:e4b-mlx, 40/44, against 11/38
+/// for a control that returns the text unchanged. A point either way is noise:
+/// the same prompt scored 37 and then 36 across two passes. The full scoreboard is at the
 /// bottom of scripts/validate-generic.py.
 ///
 /// The last two examples are a *correction* — an instruction with the
--- a/Sources/ParrotFlow/FreeForm.swift
+++ b/Sources/ParrotFlow/FreeForm.swift
@@ -15,13 +15,24 @@ import Foundation
 /// the wrong narrow tool. The router needs a separate answer, not another line
 /// in the list. See `Router.prompt(for:freeForm:)`.
 ///
-/// Scored by tests/generic-cases.yaml — 38 cases, 27 that ask for an edit and
-/// 11 that must come back untouched. On gemma4:e4b, 32/38 at 1.15s, against
-/// 11/38 for a control that returns the text unchanged. Five variants landed
-/// within one case of each other, so this wording is the model's ceiling on the
-/// task rather than the best of a wide field; the full scoreboard, including
-/// the three failures no variant fixed, is at the bottom of
-/// scripts/validate-generic.py.
+/// Scored by tests/generic-cases.yaml — 44 cases, 31 that ask for an edit and
+/// 13 that must come back untouched. On gemma4:e4b-mlx, 39/44, against 11/38
+/// for a control that returns the text unchanged. The full scoreboard is at the
+/// bottom of scripts/validate-generic.py.
+///
+/// The last two examples are a *correction* — an instruction with the
+/// imperative taken out. "make it Tuesday" and "I meant Tuesday" ask for the
+/// same edit, and the second is what people say to a machine that has just
+/// written their own sentence down. Without them this scored 3/6 on those, and
+/// it failed in one particular way: handed "no I meant three of them" it edited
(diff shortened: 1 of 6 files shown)


znat:
Routing coverage added, and it found a failure rather than a gap.

    I meant Tuesday                    ->  repetitions   ✗
    no I meant three of them           ->  ANY           ✓
    this is not what I had in mind     ->  NONE          ✓
    I meant what I said in the review  ->  NONE          ✓

On the heard path the router picks a script: `hey parrot, I meant Tuesday` runs the repetitions transform over the sentence instead of reaching the catch-all. That is worse than the three failures already standing in `tests/routing-cases.yaml`, which land on grammar — a prompt whose every rule is against touching a number, so the cost is a visible no-op rather than a rewrite.

It is recorded rather than fixed here. The fix is a transform description, and this repo changes a description against this set with a number before and after; that is not the pull request that found it. The case and the reasoning are in the file so the next person starts from the number.

The keyed path is unaffected — no router, and its four correction cases pass at 26/26.

Both complaints coming back `NONE` is correct, not a miss: the heard path has to decide whether a sentence was an edit at all, and "this is not what I had in mind" is not one. The keyed path sends it to the catch-all instead, because a key already said it was a command — and the prompt returns it unchanged, which `tests/generic-cases.yaml` scores.

PR #219, opened 2026-08-28
refactor: fold the parallel state in the edit-by-voice branch back together

znat:
Six behaviour-preserving cuts over `feat/edit-by-voice`, all in code that branch added. Nothing changes on screen, in the log, or in what any gesture does. The point is the shape: three copies of the hotkey become one, two parameters that always agreed become one, a five-field tuple becomes the struct it already was, and two near-identical offer paths become one.

Code drops 30 lines net. The line count is not the win — the win is that four places that had to be kept in sync by hand no longer can drift.

Every check in `make test` passes, including `keyed`. Review `Target.confirms` and `canRaiseAnOffer` first: those two are where a mistake would be invisible.

<details>
<summary>The hotkey was stored in three places and only ever read in one</summary>

`hotKeys.binding?.displayName` is the live source. It was copied to `PillModel.hotkey` at registration, then threaded through `PillMetrics.panelSize`, `width`, `height` and `offer` as a defaulted `hotkey: String = ""` parameter, plus two call-site arguments.

Every one of those reads sat inside a `case .selection` branch. No other pill state consulted it. So it moves into the state: `Headline.selection(String, hotkey: String)`.

That deletes `PillModel.hotkey`, the four parameters, both call-site arguments, and the two `pill.model.hotkey = …` assignments in `AppDelegate.applyConfig`. `AppDelegate.holdKey` reads the binding at the moment the headline is built.

It also fixes a small inconsistency nobody had hit. The pill is measured before it is drawn. With the key outside the state, a config reload between raising an offer and re-measuring it could change the measurement of an offer already on screen. Frozen into the headline, the two cannot disagree.

</details>

<details>
<summary>`confirm:` was exactly `!keyed` at every call site</summary>

Two parameters travelled together through `handleVoiceCommand`, `run` and `runTransform`, and never disagreed:

| call site | `keyed` | `confirm` |
|---|---|---|
| `Router.local` named it outright | from the caller | `!keyed` |
| keyed catch-all | `true` | `false` |
| router `.matched` / `.anything` | unreachable when keyed | default `true` |
| offer chip, menu | `false` | default `true` |

`Target` now carries `keyed` and answers `confirms`. `handleVoiceCommand` reads `keyed` off the target instead of taking it beside one.

The leaf `runTransform(_:instruction:selection:on:confirm:)` and `finishTransform` keep the plain bool. By then the target is resolved to a `String` and there is nothing left to ask.

`Press.frozenTarget(keyed:)` replaces the two hand-built `.frozen(selection: press.selection, fallback: press.transcript)` constructions in `deliverTranscription`.

</details>

<details>
<summary>`lastDictated` was a `Correction` written as a tuple</summary>

It was `(run: Int, text: String, element: AXUIElement?, owner: NSRunningApplication?, landing: Correction.Landing)`. Both `showCorrectOffer` and `summonOffer` unpacked it field by field to build a `Correction` from it.

It is a `Correction?` now, with `dictation` holding the run. `showCorrectOffer` builds it once and hands the same value to both `lastDictated` and `raiseOffer`. `summonOffer`'s last-dictation branch is one line.

`Correction.original` becomes `var` so `noteRewritten` can relabel the record in place. Nothing writes to an offer's own copy — that one is frozen when the offer is raised.

</details>

<details>
<summary>`summonOffer` and `offerOverReselected` shared four guards and twelve lines</summary>

Both checked `!recorder.isRecording`, `runsInFlight <= 0`, `!offerIsUp` and `config.feedback.correctOffer`, in two separately-written lists. Both then ran the same `aim(at:)` plus `raiseOffer(over: Correction(… selection …))` block.

They are `canRaiseAnOffer` and `offerOverSelection(_:)` now. `offerOverReselected` is gone; `watchForReselection`'s callback checks the guard and calls the shared path.

The two guard lists were identical today. Nothing stopped the next edit from changing one and not the other.

</details>

<details>
<summary>`runTransform` shadowed its own parameter with a different type</summary>

```swift
private func runTransform(
    _ transform: Config.Transform, instruction: String,
    progress inherited: Int?, over target: Target = .whateverIsSelected
) {
    …
    let target = selection?.text ?? dictated ?? ""   // a String, not a Target
```

Legal Swift, and it compiled. But `target` meant three things in `AppDelegate.swift` — a `Target`, a `Correction`, and this `String`. The local is `text` now.

</details>

<details>
<summary>What this deliberately leaves alone</summary>

**The comments.** They are 51% of the added lines on `feat/edit-by-voice` (825 of 1618). Trimming them to the `CLAUDE.md` standard is worth roughly four times what this PR saves, but it is a reading job, not a mechanical one, and it belongs in its own review.

**`String.isBlank` at the older call sites.** Six `trimmingCharacters(in: .whitespacesAndNewlines).isEmpty` remain in `LLM.swift`, `LocalLLM.swift`, `Config.swift`, `Template.swift` and `Transcriber.swift`. Converting them would spread this diff into five files the branch never touched. Separate sweep.

**`SelectionWatch`'s copy of the last dictation.** It holds `haystack` and `field`, re-armed after every rewrite — a third copy of `lastDictated`. A closure reading the record live would delete `watchForReselection`. But `reselect.stop()` on press closes a real gap, and moving that check inside `look()` couples the watch to the recorder. Small saving, not-small risk.

**The `keyedAtPress` rule.** It reads `offerHeadline?.isSelection`, which is UI state deciding input semantics. `offeredCorrection?.target.selection != nil` says the same thing from the data. The existing comment argues for matching the headline on purpose, so it stays.

</details>

<details>
<summary>Review notes — where a mistake would hide</summary>

Mechanical and safe:

- every `.frozen(a, b)` → `.frozen(a, b, _)` pattern update
- `last.text` → `last.original`, `last.run` → `last.dictation`
- the `target` → `text` rename in `runTransform`

Worth reading closely:

- **`Target.confirms`.** The keyed catch-all must still reach `finishTransform` with `confirm: false`, and the router branches must still get `true`. The router branches are unreachable when keyed — `handleVoiceCommand` returns before them — which is what makes the derivation safe.
- **`canRaiseAnOffer`.** Same four conditions as both old guard lists, in one place.
- **`holdKey` on a failed registration.** `register` unregisters before it tries, so `hotKeys.binding` is nil and `holdKey` is `""`, which takes the hold row off. Same as the old explicit `pill.model.hotkey = ""`.

Not run: `check-grammar`, `check-routing`, `check-spelling` (need Ollama) and `check-inplace` (needs a screen). No prompt, no routing rule and no in-place write is touched by this diff.

</details>

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -173,7 +173,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         /// Decided at the press because that is where the gesture is known, and
         /// carried here for the same reason everything else is — a second press
         /// landing mid-transcription must not be able to change what this one
-        /// was for. See `handleVoiceCommand(_:keyed:)` for what it buys.
+        /// was for. See `handleVoiceCommand(_:over:)` for what it buys.
         var keyed = false
         /// What was selected when this recording began.
         ///
@@ -195,6 +195,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         /// the instruction for it. Freezing the selection alone left this half
         /// of the same crossing open.
         var transcript: String?
+
+        /// What this press froze, as the command path wants it. `keyed` is
+        /// passed rather than read off `self.keyed` because the heard path
+        /// reaches here on a press that was not keyed and is still a command.
+        func frozenTarget(keyed: Bool) -> Target {
+            .frozen(selection: selection, fallback: transcript, keyed: keyed)
+        }
     }
 
     /// The words the offer on screen is about, and the field they went into.
@@ -215,8 +222,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// until the next dictation replaces it.
     private let reselect = SelectionWatch()
 
-    /// The last dictation a tap can summon an offer over: its words, and where
-    /// they went.
+    /// The last dictation a tap can summon an offer over.
+    ///
+    /// The same record the offer itself is raised over, kept past that offer so
+    /// a tap can raise another one from it. Its `dictation` is the run it came
+    /// from, which is never nil here.
     ///
     /// `offeredCorrection` is cleared by every ending, because it asserts that
     /// an offer is up and about these words. This asserts nothing of the sort —
@@ -231,10 +241,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// landing stays with the newer run — so a tap would have offered one
     /// dictation's words at another's field, and taking it would have rewritten
     /// there. One record, written once, cannot come apart that way.
-    private var lastDictated: (
-        run: Int, text: String, element: AXUIElement?,
-        owner: NSRunningApplication?, landing: Correction.Landing
-    )?
+    private var lastDictated: Correction?
 
     /// The focused element's text as it was at the press, for an app that gave
     /// no caret. What changed in it afterwards is where the words went.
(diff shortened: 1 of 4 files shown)


znat:
Not worth the churn — closing.

PR #222, opened 2026-08-29
fix: an ordinary first name is no longer overwritten without asking

znat:
A vocabulary term could be written over any word no dictionary knew, with
nothing reading the sentence. `NSSpellChecker` has no ordinary first names in
it, so "Um not Peter, uh Frederick." came out as "…uh Redrock." That is now
two word lists, and both have to say they have never seen the word before it
may be overwritten. Everything else goes to the judge, as it already did.

The second list is `data/wordpiece.txt` — 23694 lines, and it is most of this
diff. The code change is small: one new lookup, and one line in
`Vocabulary.autoApplies`. Start the review at
`Sources/ParrotFlow/WordPieces.swift`.

<details>
<summary>Why one word list was the wrong test — it has no first names in it</summary>

`Vocabulary.autoApplies` decides when a proposal skips the judge. For a single
word the rule was `!Replacements.isRealWord(bare)`: a word no dictionary knows
is safe to overwrite.

That test is `NSSpellChecker` over `en` then `fr`. Measured 2026-08-29 with the
shipped binary:

    unknown to it:  Sarah, Nathan, Zoe, Ahmed, Priya, Frederick, Salman,
                    Matthieu, Lucia, Bjorn, Siobhan, Xiao
    known to it:    Peter, Mira, Mathieu, Claude, subtask, repo, async,
                    yaml, rebase, tokenizer, dataset

So the rule auto-applied over ordinary names. It is not a threshold that was
set too low; it is a whole class of word the dictionary cannot see.

</details>

<details>
<summary>What the second list is, and why a tokenizer vocabulary answers this</summary>

`data/wordpiece.txt` is the whole-word half of `distilbert-base-uncased`'s
WordPiece vocabulary. A tokenizer keeps a word whole when it saw it often
enough while being trained and chops the rest into fragments, so membership is
a frequency fact rather than a lexicographer's judgement. That is what covers
the dictionary's blind spot: it knows `Sarah`, `Nathan` and `Frederick`.

Its own blind spot is the mirror image, which is the point of asking both:

| list | knows | misses |
| --- | --- | --- |
| `NSSpellChecker`, `en` then `fr` | `subtask`, `repo`, `rebase`, `tokenizer`, `Mathieu` | first names — `Sarah`, `Nathan`, `Frederick` |
| `data/wordpiece.txt` | those first names | jargon — `subtask`, `repo`, `rebase`, `tokenizer` |

**No model runs.** The whole test is a set lookup:

```swift
words.contains(word.folding(options: [.diacriticInsensitive, .caseInsensitive],
                            locale: Locale(identifier: "en_US")))
```

The tokenizer normalises by lowercasing and then dropping combining marks. I
checked the claim rather than taking it: **none of the 23694 entries carries a
combining mark**, and every one of them survives the tokenizer's own
normalisation unchanged — so the set is exactly the set of words the tokenizer
keeps whole. The folded lookup and "the tokenizer does not split this word"
agree on 43 of 43 sample words, including `Chloé` → `chloe` and `Créé` →
`cree`.

</details>

<details>
<summary>Where the file came from, and its licence — Apache-2.0, same as this repo</summary>

```sh
V=$(find ~/.cache/huggingface/hub/models--distilbert-base-uncased -name vocab.txt | head -1)
grep -v '^##' "$V" | grep -v '^\[' > data/wordpiece.txt   # 23694 lines, 180 KB
```

Snapshot `12040accade4e8a0f71eabdb258fecc2e7e948be` of
[`distilbert/distilbert-base-uncased`](https://huggingface.co/distilbert/distilbert-base-uncased).
The model card's `license` field is `apache-2.0`. This repository's `LICENSE`
is Apache-2.0, so the file ships under the same terms as the rest of the app.
No weights are shipped and none are downloaded — only the vocabulary text.

The file is copied into `Contents/Resources/wordpiece.txt` by
`scripts/build-app.sh`, named explicitly rather than by copying `data/`, whose
other files are read by `scripts/calibrate.py` on a checkout. `WordPieces.fileURL`
resolves it the way `Config.configTemplateURL` already does: the bundle when
there is one, the source tree when the binary is the raw `swift build` product
the check scripts run. Both paths verified live — with `frederick` deleted from
a bundled copy the same binary answers `auto-apply` instead of `judge`.

</details>

<details>
<summary>Measurements — 97/141 before and after on menu-recall, one wrong rewrite stopped</summary>

**`scripts/menu-recall.py --runs 3`**, the gate that must not regress. Both
runs on this machine against `.build/ParrotFlowDev.app`, warm `gemma4:e4b-mlx`:

| | recall | picked | clips that flipped between runs |
| --- | --- | --- | --- |
| before (`main`, 3ab201f) | 97/141 | 97/141 | 3 |
| after (this branch) | 97/141 | 97/141 | 1 |

Two clips swapped between the runs and cancelled out. Both were already
annotated `flipped` by the majority-of-3 in the baseline — `Versal` and
`Versailles` over `Vercel` — and neither word is one this change reroutes:
`Versal` and `Versailles` are both known to the French dictionary and already
went to the judge. That is F12a replay noise, not the change.

**`tests/judge-cases.yaml`.** Counting the proposals the word test lets
through, over the 59 cases in the file today, with the shipped `--word-gate`:

| | auto-applied | wrong |
| --- | --- | --- |
| the dictionary alone (ships today) | 18 | 6 |
| both lists | 17 | 5 |

The single case the second list stops is `Frederick.` → `Redrock`, labelled
`decline`. This count is the word test alone. It ignores the acoustic condition
`termScore > heardScore`, which cannot be replayed without the audio. An earlier
measurement that included that condition covered 50 of these cases and put the
same change at 15 auto-applied with 2 wrong, becoming 14 with 1 wrong. Either
way the delta is one case, and it is the same case.

**Be careful how much this is.** The labels are hand-written, from one
speaker's archive — 50 cases in the measurement with the score condition. The
change stops one wrong rewrite and costs one auto-application. The argument for
it is not that it moves a score. It is that it removes a class of failure,
overwriting ordinary first names, that no threshold could have reached.

**`make test`**: every check passed, including the new one.

</details>

<details>
<summary>The new set — 19 cases, plus the fail-open case</summary>

`scripts/check-word-gate.sh` against `tests/word-gate-cases.yaml`, run by
`make test` and by CI. It scores the shipped lists through a new
`--word-gate <word>`, which prints both verdicts and the decision:

    $ ParrotFlow --word-gate Frederick
    spell      unknown
    wordpiece  known
    gate       judge

Both verdicts are checked and not only the decision. A word reaches `judge`
from either side, so a set reading the decision alone would pass with one half
broken — `Chloé` in particular, where the accent is the thing under test.

The four cases the change is about:

| word | spell | wordpiece | gate |
| --- | --- | --- | --- |
| `Frederick` | unknown | known | judge |
| `subtask` | known | unknown | judge |
| `Versal` | unknown | unknown | auto-apply |
| `Chloé` | unknown | known | judge |

**Fail open.** `WordPieces.knows` returns `Bool?` and not `Bool`, and the third
answer is the point. A missing file that read as "unknown" would put the gate
back exactly where it was, silently, on any machine where the resource did not
ship. The script points `PARROTFLOW_WORDPIECE` at a path that does not exist
and requires `Versal` — the one word that otherwise auto-applies — to come back
`judge`.

Half the check asks `NSSpellChecker`, which intermittently times out and reads
as "known word" when it does. It happened once while this was being written:
`Versal` came back `known` under load, then `unknown` ten times in a row. So a
failing case is asked a second time in a fresh process — `isRealWord` caches
per process, so that is a real second sample — and only then called failed. A
regression fails both times.

</details>

<details>

- Names both lists miss still auto-apply: `Priya`, `Siobhan`, `Matthieu`. Two
  cases of this are in the set, recorded rather than hidden.
- Compounds both lists miss still auto-apply: `webhook`, `worktree`,
  `monorepo`, `kubernetes`. The briefing for this change assumed the dictionary
  knew these. Measured, it does not. They behave exactly as they do today.
- `Mirza's` → `Mirza` is still wrong in `tests/judge-cases.yaml`. That is the
  possessive path, not this gate.
- The judge, its prompt, slot-POS and word-rank are all untouched.

</details>

<details>
<summary>Review notes — the code change is small, the resource is not</summary>

`data/wordpiece.txt` is 23694 lines and dominates the diff. Everything else:

| file | what |
| --- | --- |
| `Sources/ParrotFlow/WordPieces.swift` | new. The list, the fold, the three-valued answer, the path. Start here. |
| `Sources/ParrotFlow/Vocabulary.swift` | `autoApplies` now calls a new `unseenWord`, which is the word half of the gate. Plus the doc comment above it. |
| `Sources/ParrotFlow/WordGateCommand.swift` | new. `--word-gate` calls `Vocabulary.unseenWord` for the decision, so the set scores the shipped test and not a copy of it. |
| `scripts/build-app.sh` | one `cp` into the bundle. |
| `scripts/check-word-gate.sh`, `tests/word-gate-cases.yaml` | new set. |
| `Makefile`, `.github/workflows/checks.yml` | register the check in both, kept in step. |
| `docs/transcription.md`, `docs/cli.md`, `docs/authoring.md` | the rule as it now reads. |

The split-compound branch at the top of `autoApplies` is untouched. It returns
before the word test, deliberately — "red crawl" for `Redcrawl` is a question
about spelling, not about whether the word exists.

</details>

- [x] `make test` passes.
- [x] A pattern change is scored, and the numbers are above.
- [x] Every commit is signed off — `git commit -s`.


znat:
Answering the one thing T-Rex could not do: the native run and the bundle
inspection. Both were done on macOS 27, Apple silicon, before this was opened.

**`scripts/check-word-gate.sh`** — 19/19 plus the fail-open case, against
`.build/release/ParrotFlow`. It is in `make test` and in
`.github/workflows/checks.yml`, and the `checks` job on this PR is green, so
the runner has run it too.

**The bundle.** `WordPieces.fileURL` has two arms and the source-tree arm would
hide a broken bundle arm, because both files have the same content. So the
bundle arm was proved by making the two differ. `frederick` deleted from a
copy of the bundled list, same binary:

    $ cp -R .build/ParrotFlowDev.app /tmp/pf-bundle-test/
    $ grep -v '^frederick$' <bundled list> > /tmp/.../Contents/Resources/wordpiece.txt

    $ /tmp/pf-bundle-test/ParrotFlowDev.app/Contents/MacOS/ParrotFlow --word-gate Frederick
    spell      unknown
    wordpiece  unknown
    gate       auto-apply

    $ .build/ParrotFlowDev.app/Contents/MacOS/ParrotFlow --word-gate Frederick
    spell      unknown
    wordpiece  known
    gate       judge

The answer follows the file, so the installed app reads
`Contents/Resources/wordpiece.txt` and the build-directory binary reads the
source tree. Both arms are live.

**`scripts/menu-recall.py --runs 3`**, the gate that must not regress: 97/141
recall and 97/141 picked, before and after. Detail in the measurements block
above.

PR #226, opened 2026-08-29
feat: ask the sentence model what word belongs in a slot

znat:
The model #225 downloads now has a reader. This adds a byte-level BPE tokenizer, a masked-word probe over it, and `--sentence-probe` to run one by hand. Nothing in the app calls it yet — two stages will, and this is the shape they need. Start the review at `SentenceProbe.at(left:right:)`, which is the whole API, and at `BPETokenizer.assertBoundaryIDs`, which is what stops the leading-space mistake shipping.

The acceptance test for the model itself:

```
$ ParrotFlow --sentence-probe "The capital of Ireland is [MASK]."
  boundary  "." 15   " ." 964   "That" 2773   " That" 2064   ✓
  mask at 6 of 64
   1   Dublin           -0.19
   2   Belfast          -3.02
   3   Cork             -3.16
   4   London           -4.10
   5   Edinburgh        -4.59
   6   Liverpool        -5.01
   7   Glasgow          -5.33
   8   Oxford           -5.56
   9   Manchester       -6.43
  10   Birmingham       -6.46
  one forward pass, 8 ms warm, 18 ms on the first
```

`Dublin, Belfast, Cork, London` is the correct list. A conversion that only reads one token answers `£, isation, organisation, ised` instead. That is what `znaat/modernbert-coreml` was published to rule out, and it is also what a wrong tokenizer or wrong input packing produces, so this one command covers the model, the BPE and the packing at once.

<details>
<summary>The leading space is the trap, and this is how it is caught</summary>

On a byte-level BPE a word carries the space in front of it, and the period does not:

```
"we have to do it. That works well"
   ->  we Ġhave Ġto Ġdo Ġit . ĠThat Ġworks Ġwell
```

So the id for a word that follows another word is `encode(" That")[0]`, which is 2064. `encode("That")[0]` is 2773 and is a different token. A caller that compares against 2773 finds it at the top of nothing and scores zero. It errors on nothing and looks like a model problem.

`BPETokenizer.assertBoundaryIDs` runs at load. It tokenizes `"we have to do it. That works well"` for real and checks four things against that tokenization, not against remembered numbers:

- the token at the period's position is `firstID(of: ".")`
- the token after it is `firstID(of: " That")`
- `" That"` and `"That"` are different ids, and so are `"."` and `" ."`
- both ids read back as `"."` and `" That"` through `word(of:)`

The last one covers the id-to-word direction, which the top-k caller depends on and which nothing else exercises. Loading throws if any of it fails, so a broken tokenizer cannot reach a measurement.

The CLI prints the same four ids on its `boundary` line, so the check is visible in any run.

ModernBERT has two period tokens, `.` (15) and `Ġ.` (964). A caller asking "is there a period in this slot" has to take the better of the two. `--against` does, and says so.

</details>

<details>
<summary>The API, and the two callers it was shaped for</summary>

One call is one forward pass. Nothing batches, on purpose: a caller that ranks the windows of a sentence pays 8 ms per window and should see that in its own code.

```swift
let probe = try await SentenceProbe.load()
let slot = try probe.at(left: "we have to do it", right: " works well")

slot.logProbability(of: id)   // one named token
slot.top(10)                  // readable words, leading space kept
```

**The sentence join.** Is a period real, or did a pause cut a sentence in two? One forward pass, then compare the next word against the period.

```
$ ParrotFlow --sentence-probe "the report is due on [MASK] morning" --against " Friday"
  " Friday"         -2.43
  period            -7.71

$ ParrotFlow --sentence-probe "we have to do it [MASK] works well" --against " That"
  " That"           -8.20
  period            -4.35
```

The first cut was spurious. The second period is real. The probe separates them by 5.3 and 3.9 nats.

**The vocabulary gate.** `top(10)` returns words, so the caller can put each back in the sentence and tag it with `NLTagger`. Rank comes from `logProbability(of:)` on the span's own token, one `at` call per window.

`left` and `right` are literal text and are passed to the tokenizer byte for byte. The mask stands for the token including its leading space. So `left` ends on a word with no trailing space, and `right` carries its own leading space when the next thing is a word — `"."` when it is punctuation. The doc comment on `SentenceProbe` says this first.

</details>

<details>
<summary>The BPE matches a reference implementation on 14 of 14, including CJK and emoji</summary>

There is no reference tokenizer on this machine and no package was added, so the Swift was checked against a Python BPE written from the same `tokenizer.json`. Exact id sequences, not counts:

| case | |
|---|---|
| `The capital of Ireland is` | ok |
| `we have to do it. That works well` | ok |
| `Naïve café résumé, don't you think` | ok |
| `Send the webhook to Mirza's endpoint at 3:45pm` | ok |
| `ParrotFlow—the dictation app—works offline (mostly)` | ok |
| `  double  spaces   here` | ok |
| `日本語のテキストです` | ok |
| `emoji 🚀 and fi ligature` | ok |
| `CamelCaseIdentifier snake_case_name kebab-case` | ok |
| `It cost $1,234.56 on 2026-08-30 at 07:15` | ok |
| `"quoted" 'single' \`backtick\` <angle> {brace}` | ok |
| `a` / ` a` | ok |
| a 28-word sentence | ok |

The underscore case caught a bug in the Python, not in the Swift. Python's `\w` includes `_`, so a `[^\s\w]+` punctuation class drops underscores silently. The Swift uses ICU `\p{L}` and `\p{N}`, which is what HuggingFace's ByteLevel pre-tokenizer means, and it was right.

</details>

<details>
<summary>The window is ±12 words, and it is in this PR so both callers get it by default</summary>

Measured earlier on 200 real periods and 120 cuts:

| window | fits in 64 | at threshold −2 | at 0 |
|---|---|---|---|
| whole sentence | 89% | 56% / 3.0 false | 82% / 7.5 |
| ±24 words | 99% | 54% / 1.5 | 82% / 6.0 |
| ±12 words | 100% | 56% / 0.5 | 82% / 3.0 |

Full coverage and a better signal. A word twelve away is noise, not context.

The trim slices the original string at a space and keeps the end that touches the mask. It does not split and rejoin: that would eat the leading space `right` needs, which is the same trap as above, reintroduced by the helper meant to be safe.

Verified: 16 words either side trims to exactly 12 and 12, `right` keeps its leading space, `"."` on the right passes through untouched, 12 words either side is left alone, 13 trims to 12.

Past the window there is a token budget guard for the input nobody measured — a long run of rare words, or a caller that skipped the window. It drops from the far ends and never from the mask. With 12 words per side that each cost 4 tokens it trims to 30/31 and still answers.

</details>

<details>
<summary>Reading one logits row in place, not converting the array — 450 ms to 8 ms</summary>

The output is `logits [1, 64, 50368]` and only the mask's row is wanted. The first version read it with `MLShapedArray<Float>(converting:)`, which is correct and walks all 3.2 million values. A call cost 450 ms.

Reading the one row through `withUnsafeBufferPointer(ofType:)` costs 8 ms warm, which matches what was measured for this model. `dataType` is checked for `.float32`, `.float16` and `.double`, and the strides are checked to be contiguous rows first, so the pointer read cannot silently return the wrong values if the conversion ever changes shape.

`logSumExp` is computed once per `Slot`, not per `logProbability` call.

The tokenizer is parsed once, on `SentenceModel`, beside the model. Building the two 50k tables out of the 2 MB `tokenizer.json` costs 132 ms, and a caller that loaded a probe per sentence would have paid that per sentence.

</details>

<details>
<summary>Checks, and what is not checked</summary>

`swift build` is clean. `make test` passes every check. SwiftLint reports nothing on the three new files.

There is no new script in `CHECKS`. Everything in that list runs with no model, and CI has no 300 MB download. A check that needed `tokenizer.json` would either go red on CI or skip silently, and a check that always skips is worse than none. The load-time assertion and the `boundary` line in the CLI are what carry it instead.

`--sentence-probe` is documented in `docs/cli.md`, under "What word belongs in a slot".

608 lines across the three new files.

</details>

### DIFF
--- a/Sources/ParrotFlow/BPETokenizer.swift
+++ b/Sources/ParrotFlow/BPETokenizer.swift
@@ -1,0 +1,295 @@
+import Foundation
+
+/// The byte-level BPE behind ModernBERT, read from a HuggingFace `tokenizer.json`.
+///
+/// Ours rather than swift-transformers. One model, one tokenizer file, no
+/// package dependency: the vocabulary, the merge ranks, the GPT-2 byte
+/// alphabet and the four special ids are all in that one file.
+///
+/// The thing to know before using it — and the thing that silently scored 0%
+/// when it was got wrong — is that a word carries its leading space:
+///
+///     "we have to do it. That works well"
+///        -> we Ġhave Ġto Ġdo Ġit . ĠThat Ġworks Ġwell
+///
+/// So the id for a word that follows another word is `firstID(of: " That")`,
+/// 2064, and not `firstID(of: "That")`, 2773. The period after `it` is the
+/// bare `.`, 15, while a period that follows a space is `Ġ.`, 964. A caller
+/// asking "is there a period here" has to take the better of the two.
+/// `assertBoundaryIDs` checks all of that against a real tokenization at load.
+struct BPETokenizer {
+
+    enum Failure: LocalizedError {
+        case unreadable(String)
+        case boundary(String)
+
+        var errorDescription: String? {
+            switch self {
+            case .unreadable(let what): return "tokenizer.json: \(what)"
+            case .boundary(let what): return "tokenizer.json: the boundary check failed, \(what)"
+            }
+        }
+    }
+
+    private let vocabulary: [String: Int]
+    /// Merge rank by pair. Lower wins, which is the order the file lists them in.
+    private let ranks: [Pair: Int]
+    private let words: [String]
+
+    let cls: Int
+    let sep: Int
+    let mask: Int
+    let pad: Int
+
+    /// 50368 — the vocabulary plus the added tokens, and the width of a
+    /// logits row.
+    var count: Int { words.count }
+
+    private struct Pair: Hashable {
+        let left: String
+        let right: String
+    }
+
+    // MARK: - Loading
+
+    static func load(contentsOf url: URL) throws -> BPETokenizer {
+        let raw = try Data(contentsOf: url)
+        guard let root = try JSONSerialization.jsonObject(with: raw) as? [String: Any] else {
+            throw Failure.unreadable("not an object")
+        }
+        guard let model = root["model"] as? [String: Any],
+              let vocabulary = model["vocab"] as? [String: Int] else {
+            throw Failure.unreadable("no model.vocab")
+        }
+        guard let merges = model["merges"] as? [Any] else {
+            throw Failure.unreadable("no model.merges")
+        }
+
+        var ranks: [Pair: Int] = [:]
+        ranks.reserveCapacity(merges.count)
+        for (rank, entry) in merges.enumerated() {
+            // Written "Ġ t" in the version this file uses, ["Ġ", "t"] in the
+            // newer one. Both are in the wild for ModernBERT.
+            let halves: [String]
+            if let line = entry as? String {
+                halves = line.components(separatedBy: " ")
+            } else if let pair = entry as? [String] {
+                halves = pair
+            } else {
+                throw Failure.unreadable("merge \(rank) is neither text nor a pair")
+            }
+            guard halves.count == 2 else {
+                throw Failure.unreadable("merge \(rank) has \(halves.count) halves")
+            }
+            ranks[Pair(left: halves[0], right: halves[1])] = rank
+        }
+
+        // The added tokens sit above the vocabulary — 50280 to 50367 here —
+        // and a prediction can land on any of them, so the readable side has
+        // to cover them or `word(of:)` returns nothing for a real answer.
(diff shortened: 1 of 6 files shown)


znat:
Closing in favour of #227, which covers the same ground with tests.

Both add `BPETokenizer.swift`, `SentenceProbe.swift` and `SentenceProbeCommand.swift` on top of main, so they conflict on every file. #227 ships the same Swift (622 lines against 661 here) plus a tokenizer fixture checked against HuggingFace's own output, a score reference against the Python probe, two check scripts and the CI wiring. This branch has no tests.

This PR was opened by a stale agent that resumed after being stopped. It was not asked for. The branch is kept if anything here is worth lifting.

PR #227, opened 2026-08-29
feat: the sentence model says whether a period is real

znat:
A pause in the middle of a sentence makes the transcriber write a period. This
teaches the model that #225 downloads to say whether that period is real. Put a
`[MASK]` where the period is, lowercase the word after it, and read both
candidates at one masked position: `score = log P(".") − log P(" <next word>")`.
Below the threshold the period is false and the two halves are one sentence.

Nothing in the app calls it yet. That is the PR after this one. What lands here
is a tokenizer, the probe, and `--sentence-probe`.

Check the tokenizer first. It is hand-written, no swift-transformers, and its
mistakes are silent — the trap below cost a whole run and raised no error. Two
scripts guard it: `scripts/check-tokenizer.sh` scores 19 cases against
HuggingFace's own answers and is in `make test`;
`scripts/check-sentence-probe.sh` compares 40 real boundary scores against
coremltools on the same `.mlpackage`, and is not, because it needs the 300 MB
model.

```
$PF --sentence-probe "we have to do it." "That works well"
  boundary  "." 15   " ." 964   "That" 2773   " That" 2064  ✓
  text      we have to do it [MASK] that works well
  score        1.45
  period      -4.12   "."
  next        -5.57   " that"
  top       " and"   -1.54   ","   -2.36   " something"   -2.93
```

<details>
<summary>The trap: the word after a boundary carries a space, the period does not</summary>

In this tokenizer a word that follows another word carries its leading space:

```
"we have to do it. That works well"  ->  we Ġhave Ġto Ġdo Ġit . ĠThat Ġworks Ġwell
```

So the period is the bare `.`, 15, and the next word is the first token of
`" " + word`, 2064 — not of `word`, 2773, a rare no-space token that almost
never occurs in text.

Taken naively the probe scores **0% repaired at every threshold** and nothing
crashes. It was measured. It only looks correct on a WordPiece tokenizer, which
is where the original script was written.

`BPETokenizer.assertBoundaryIDs` checks both ids against a real tokenization of
`"we have to do it. That works well"` at load, rather than hard-coding numbers.
`load` throws if it fails, so the probe refuses to answer instead of returning
one. Verified by hand with a tokenizer whose merges were emptied:

```
✗ tokenizer.json: "we have to do it. That works well" is 33 ids, want 9
```

The four ids are printed on the `boundary` line of every run, because a check
you cannot see is a check you stop trusting.

</details>

<details>
<summary>Thresholds — 32% of cuts repaired at 0.0 false joins per 100 real periods</summary>

Measured over 194 hand-labelled real periods and 130 synthetic cuts of English
dictation. **repaired** is the share of cuts the probe removes, higher is
better. **false joins** is genuine periods removed per 100, lower is better.

| threshold | cuts repaired | false joins per 100 real periods |
|---:|---:|---:|
| −4 | 32% | 0.0 |
| −2 | 55% | 1.2 |
| 0 | 82% | 6.1 |

No threshold is wired into anything. The PR that calls the probe picks one.

The window is ±12 words each side, the next word counting as the first on the
right. Fidelity of this conversion falls off past sequence length 64
(correlation against PyTorch at the mask is 0.9964–1.0000 at 64 and
0.8996–0.9862 at 128), and a word twelve away measured as noise rather than
context.

</details>

<details>
<summary>Swift against Python — 40/40 within 0.05, worst disagreement 0.0050</summary>

A hand-written BPE goes wrong in ways no single example catches, so the check
was built before the feature.

**Tokens.** `tests/tokenizer-cases.json` holds what HuggingFace's `tokenizers`
answers for 19 cases: plain sentences, punctuation, apostrophes, accents,
digits, capitals, multiple spaces, an emoji, CJK, `[MASK]` inside the text, the
empty string, `"a "` and `" a"`, the four boundary tokens, and one case of 156
tokens so the over-length path is exercised.
`scripts/check-tokenizer.sh` compares `--sentence-probe --encode` to it, id by
id. **19/19.** It runs on `tokenizer.json` alone and uses the cached copy if
there is one, otherwise fetches 2 MB from the pinned revision — a run that
cannot get the file fails rather than passing by skipping.

Two cases would have failed a simpler tokenizer:

- `"  double  spaces   here"` → `50276 12237 50276 31748 50275 1568`. Runs of 2
  to 24 spaces are **added tokens**, matched longest-first before the BPE runs.
- `"we have to do it [MASK] that works well"` → `[MASK]` has `lstrip: true`, so
  it eats the space in front of it and the word after starts bare.

**Scores.** `tests/sentence-boundary-cases.json` holds 40 boundaries — 20 real
periods and 20 synthetic cuts, taken from the scored set — with the score, and
both log probabilities, that coremltools gets from the same `.mlpackage` Swift
loads. Same weights, same runtime, so a disagreement is Swift's tokenization or
packing and nothing else.

```
scripts/check-sentence-probe.sh
  40/40 within 0.05  (tests/sentence-boundary-cases.json)
  worst gap 0.0050  "pinned row could clear a newer."
```

The tolerance is 0.05 and the largest disagreement was **0.0050**, which is the
size of float16 rounding. It was not widened to make anything pass.

PyTorch was **not** used as the reference. `run.py`, which produced the recorded
numbers, runs PyTorch unpadded at length 384. Swift runs Core ML padded to 64.
Comparing against PyTorch would fold the conversion's own error into the number
and invite widening the tolerance, when the thing under test is Swift's packing.

Padding matters: `attention_mask` is not an input to this conversion — the
wrapper builds it as `ones_like(input_ids)`, so the padding is attended. The
reference pads with `[PAD]` exactly as Swift does.

</details>

<details>
<summary>Review notes — where to start, and what the fixtures contain</summary>

Start at `BPETokenizer.assertBoundaryIDs`, then `SentenceProbe.read`. The rest
is mechanical.

`BPETokenizer.swift` is 316 lines and the only new logic worth reading closely:
the added-token scan in `encode`, the GPT-2 byte alphabet, and the merge loop.
It is parsed once per process and cached on the `SentenceModel` actor: reading
the file and building its two 50k tables costs about 90 ms, against 8 ms for
the forward pass.
`SentenceProbe.Slot` reads one row of the logits in place — going through
`MLShapedArray<Float>` first is correct and costs 440 ms of the 450 a call took,
because it walks all 64 × 50368 values to reach the one row.

`--sentence-probe --encode` exists for the check script: it loads the tokenizer
and no model, so it answers on a machine that has never dictated.
`PARROTFLOW_TOKENIZER` points it at a copy.

`tests/sentence-boundary-cases.json` carries 40 short fragments of the author's
own dictation, all of it ParrotFlow development talk. It was read; there is
nothing personal in it. It is the same corpus the thresholds were measured on.

Not covered: French. The same probe scores AUC 0.88 in French with XLM-R and
near chance with the two published French ModernBERTs, and the French real-period
set is 51 boundaries, too few to fit a threshold. English only for now.

</details>


znat:
Two things from #226 worth lifting before this merges. Both are in the closed branch `feat/sentence-probe`, commit `a45f63c`.

**1. `Float16` needs a macOS 15 guard.** `SentenceProbe` is `@available(macOS 14, *)`, but `Float16` conforms to `MLShapedArrayScalar` only from macOS 15, so `withUnsafeBufferPointer(ofType: Float16.self)` builds with:

```
warning: conformance of 'Float16' to 'MLShapedArrayScalar' is only available in
macOS 15.0 or newer; this is an error in the Swift 6 language mode
```

It is a warning today and a build error later. The fix:

```swift
case .float16:
    // `Float16` conforms to `MLShapedArrayScalar` only from macOS 15.
    guard #available(macOS 15, *) else {
        throw Failure.shape("float16 logits need macOS 15")
    }
    self.logits = array.withUnsafeBufferPointer(ofType: Float16.self) { ... }
```

**2. `tokenizer.json` is parsed on every `load()`.** `SentenceProbe.load` calls `BPETokenizer.load(contentsOf:)` each time. Measured here: 132 ms to parse the 2 MB file and build the two 50k tables, against 8 ms for the forward pass the caller wanted. `SentenceModel.shared.prepare()` caches the model, so this is the only part that repeats.

A caller that loads a probe per sentence pays it per sentence. Caching it on the `SentenceModel` actor is about 10 lines:

```swift
private var tokenizer: BPETokenizer?

func loadTokenizer() throws -> BPETokenizer {
    if let tokenizer { return tokenizer }
    let loaded = try BPETokenizer.load(contentsOf: Self.tokenizerURL)
    tokenizer = loaded
    return loaded
}
```

That does not fit this PR's `tokenizerURL` override for `PARROTFLOW_TOKENIZER`, so it needs adapting — the cache key would have to be the URL, or the override stays uncached for the check script.

Not worth lifting: the rest of #226 is the same design, and `check-tokenizer.sh` here is better than what I did, which was to skip CI coverage entirely.

znat:
Both taken, in `12124dd`.

**1. `Float16`.** Guarded with `#available(macOS 15, *)`, and the branch throws
`float16 logits need macOS 15` rather than falling through to a wrong number.
The build is warning-free on those lines now.

**2. The parse.** Cached on the `SentenceModel` actor, beside the compiled
model, keyed by URL so `PARROTFLOW_TOKENIZER` still reaches its own copy:

```swift
private var tokenizers: [URL: BPETokenizer] = [:]

func tokenizer(at url: URL) throws -> BPETokenizer {
    if let cached = tokenizers[url] { return cached }
    let loaded = try BPETokenizer.load(contentsOf: url)
    tokenizers[url] = loaded
    return loaded
}
```

Measured here rather than taking the number: `--version` costs 63 ms per
process and `--sentence-probe --encode "hi"` costs 150 ms, so the read, the
parse and the two 50k tables are about 90 ms. Same order as your 132 ms, and
either way it is the whole budget against an 8 ms forward pass.

`--sentence-probe --encode` still calls `BPETokenizer.load` directly. That
command is a process that parses once and exits, and it must not touch the
actor that owns the 300 MB model.

`BPETokenizer` is now `Sendable` so it can leave the actor.

Re-ran after both changes: `make test` green, `scripts/check-tokenizer.sh`
19/19, `scripts/check-sentence-probe.sh` 40/40 within 0.05 with the worst gap
still 0.0050.

znat:
Merged at `4248cee`, so three commits pushed after that did not land. They are in #228: the tokenizer parse cache you asked for, the `Float16` macOS 15 guard, and an off-by-one in the window found on re-reading. Nothing else changed, and the score check is still 40/40 within 0.05 with the worst gap 0.0050.

PR #237, opened 2026-08-30
fix: `review: false` stops the model and nothing else

znat:
`review: false` returned before the stage had gathered anything, so it also
turned off the near-miss pass, the sound pass, and the free rules in front of
the judge. The reason written above it was that nothing those propose is ever
written without a model reading the sentence first.

That was true until the gate existed. It is not true now: the two word lists
write a name with no model at all, measured at no errors over one speaker's
archive, and refusing to run them because the model is off throws away the half
of the stage that never needed one.

So the guard moves below the gathering and below the gate. What the free rules
settle is written, what they do not keeps what is already there — the same
three-way answer every other no-model path in this stage gives.

```
- stage: vocabulary
  review: false
```

```
"I opened Ghost E this morning"  ->  "I opened Ghostty this morning"
vocabulary gate: 1 of 1 settled without asking
vocabulary.asked = 0
```

## The two cases that moved

`tests/pipeline-cases.yaml` had `slots: 0` on both `review-off` cases and they
are now `slots: 1`. The text each asserts on is unchanged.

That is the point rather than a concession. The stage used to return before it
had counted anything, so `review: false` reported an empty sentence whatever
was in it. "One place was found" and "nothing was written in it" are two
different facts and only one of them used to be visible.

## Testing

`make test` green.

https://claude.ai/code/session_01NMVgtb4jmhQtvWtSJr1bce

### DIFF
--- a/Sources/ParrotFlow/Pipeline.swift
+++ b/Sources/ParrotFlow/Pipeline.swift
@@ -1006,14 +1006,6 @@ struct Pipeline: Equatable, Codable {
             return StageResult(text: text, vars: wrote.merging(vars) { _, new in new })
         }
 
-        // `review: false`. The exact matches ship as the rules wrote them, and
-        // the near-miss pass is skipped with the review rather than left to
-        // run: nothing it proposes is ever written without a model reading the
-        // sentence first.
-        guard step.reviewEnabled ?? true else {
-            return result(text, ["asked": .int(0), "slots": .int(0)])
-        }
-
         let caps = step.caps ?? VocabularyJudge.Caps.standard
         // Both sources. The acoustic pass proposes with positions; a
         // `replacements` rule publishes none, so its substitutions are found by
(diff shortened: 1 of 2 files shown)


znat:
The write is real and the cause is not. Measured on the branch, same fixture, same sentence, only `review:` differing:

```
review: gemma  ->  The Redcrawl finished the run.
review: false  ->  The Redcrawl finished the run.

vocabulary gate: "Red Cross" -> "Redcrawl" is the worst-reading spot of its
                 sentence (rank 0 of 5) — written, not asked
vocabulary.asked = 0
```

The reviewer does not run on that place when it is configured either. `judgeVocabulary` asks the model only about the places the gate left open — `asked = asking.map { changes[$0] }`, and `asking` excludes everything `settle` decided — so a rank-based apply has never reached a model in any configuration. "Despite no reviewer running" describes both columns.

The proposed fix would therefore make `review: false` *stricter* than `review: gemma`: the same sentence would keep `Red Cross` with the model off and lose it with the model on. That inverts what the setting means, and it would hide the real problem rather than fix it.

The real problem is the rank rule itself, and it is the same with or without a reviewer. It asks whether a word is *unexpected* where the question is whether it is *wrong*, and a rare correct noun is both. Four live examples in one evening: `Versailles Castle -> Vercel Castle`, `Red Cross -> Redcrawl`, `mic -> Mik`, and one refusal in the other direction, `ghosty -> Ghostty` thrown away because the spot wanted a verb. It has its own switch for exactly this reason:

```yaml
gate_rank: false
```

which keeps the three rules that ask absolute questions — is this a word any dictionary knows, can a name stand in this spot, does it sound exactly like the term — and drops the one that compares a word with the rest of its sentence. On this speaker's archive the rank buys 7 correct writes and costs 1; measured over one evening of real use it cost 4.

The `Versailles Castle` case in the review is a different mechanism again, and it predates this PR: there the substitution comes from an exact rule, which has already rewritten the text, so with no reviewer it ships. `origin/main` does the same thing:

```
origin/main, review: false  ->  I want to visit the Vercel Castle.
```

So this branch does not change that one either way.

Leaving the code as it is. Happy to be argued out of it if the measurement above is read differently.

PR #254, opened 2026-09-02
fix: a word added beside another is not a correction of it

znat:
Typing a name after a word, or pasting into the line, was offered as a vocabulary rule. Both happened in one dictation: `morning` → `morning Tasmeen`, and `dictation.` → `dictation.[Image #17]`. Neither is a correction of anything — the first adds a name after a word that was heard correctly, and the second is a paste.

The panel now never sees them. `EditWatch.added` refuses a pair where one reading stands whole at the start or end of the other with a separate word beside it.

Reviewers: the whole change is `EditWatch.added` and the one `guard` that calls it. The interesting part is what it must **not** refuse.

<details>
<summary>Why "one is a prefix of the other" is the wrong test</summary>

`Ghost` → `Ghostty` is a real correction and `Ghostty` starts with `Ghost`. So a prefix test would throw away the commonest kind of name fix there is.

What separates an addition from a growth is whether a **whole separate word** appeared. So the longer reading has to hold a space before this fires at all, and what is left after removing the shorter one has to hold whitespace.

```
  morning.     -> morning Tasmeen.       rest is " Tasmeen"      refused
  dictation.   -> dictation.[Image #17]  rest has a space        refused
  Ghost        -> Ghostty                no space anywhere       kept
  Praisy       -> Praisy's               no space anywhere       kept
  Ghost E.     -> Ghostty.               not a prefix either way kept
```

</details>

<details>
<summary>Why it is asked before the trim</summary>

`added` takes the two readings **as they stood**, not what `trimmed` hands back, and takes off only the punctuation both end on.

#259 landed `prone.maybe` → `prone point.maybe`, which has to read as `prone -> prone point` and is turned away later, at the offer, as ordinary English. That is a word added beside another word, exactly like `morning` → `morning Tasmeen`. On the trimmed readings the two are the same shape and `added` cannot tell them apart.

On the readings as they stood it can. `trimmed` cuts the shared `.maybe` — the next sentence, not the correction — and only after that cut does `prone point.maybe` look like a word appended. Before it, `prone.maybe` stands at neither end of `prone point.maybe`, because the word went *inside* the run.

</details>

<details>
<summary>Why the paste looked like a correction at all</summary>

`EditWatch` splits a line on whitespace. `[Image #17]` pasted straight after `dictation.` with no space between them, so the word `dictation.` became the word `dictation.[Image` — one word for one, which is exactly the shape of a mangled name.

It is caught because the pasted text carries its own space, so the pair as a whole holds one.

A paste with no space in it at all would still get through. That is a narrower hole than the one being closed and it is not fixed here.

</details>

<details>
<summary>Five cases, three of which are the ones that must survive</summary>

```
  ✓ no single change      "…this morning."     -> "…this morning Tasmeen."
  ✓ no single change      "said morning, and"  -> "said morning Tasmeen, and"
  ✓ no single change      "…the dictation."    -> "…the dictation.[Image #17]"
  ✓ Ghost -> Ghostty      the word grows
  ✓ Praisy -> Praisy's    a possessive
```

`Ghost E. -> Ghostty.`, `prone -> prone point` and the rest of the set still pass — 39 of 39 in `tests/edit-diff-cases.yaml`.

`scripts/check-edit-diff.sh` is not one of the steps in `.github/workflows/checks.yml`, so the green tick on this pull request is the build, SwiftLint and the other sets, not this case file. Run `make test` to score it.

</details>

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -259,3 +259,10 @@ jobs:
       # scripts/check-sentence-join.sh, is run by hand.
       - name: The capital after a period the join removes
         run: scripts/check-sentence-case.sh
+
+      # Endings Parakeet writes over silence. Word arrays out of trace.jsonl,
+      # so no audio and no model — and the half that can go wrong quietly is
+      # scored too: the text is a separate string from the token timings, and
+      # a run dropped from one and not the other is invisible.
+      - name: Endings the decoder invented
+        run: scripts/check-invented-tail.sh
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,8 @@ release-certificate:
 CHECKS := numbers replacements pipeline pipeline-config wake split dotted dates keyed \
           transform-folders eval audio-recovery possessive word-gate suggest input join \
           profiles span-rule clipboard default-config vocabulary-config learn \
-          signing-identity no-voice sound tokenizer sentence-case term-uses edit-diff sentence-open
+          signing-identity no-voice sound tokenizer sentence-case term-uses edit-diff sentence-open \
+          invented-tail
 
 test:
 	@swift build -c release
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -3795,11 +3795,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         let ours = heard.contains { $0.caseInsensitiveCompare(back) == .orderedSame }
         do {
             try TermUses.record(term: term, said: sentence, span: back, counter: true)
+            rebuildPortrait(for: term)
             // One term corrected into another — `Praizy` into `Praisy` — says
             // two things at once, and both are worth keeping: the first does
             // not live here and the second does.
             if let right = existingTerm(named: back) {
                 try TermUses.record(term: right, said: sentence, span: back)
+                rebuildPortrait(for: right)
             }
             Log.write(
                 "correction: \"\(written)\" -> \"\(back)\" is a counter-example for"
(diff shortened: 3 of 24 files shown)


znat:
`main` merged in. One of the two conflicts was real and changed `added`, so it is worth a read.

#259 landed the case `prone.maybe` → `prone point.maybe`, expected to read as `prone -> prone point` and to be turned away later, at the offer, as ordinary English. That is a word added beside another word, and so is `morning` → `morning Tasmeen`. On the readings `trimmed` hands back, the two are the same string shape and `added` refused both.

On the readings **as they stood** they are not the same. `trimmed` cuts the shared `.maybe` — the next sentence, not the correction — and only after that cut does `prone point.maybe` look like a word appended to `prone`. Before it, the short reading stands at neither end of the long one, because the word went *inside* the run.

So `added` is now asked before the trim, and takes off only the punctuation both readings end on.

<details>
<summary>The five cases, and the one that made the difference</summary>

```
  morning.     -> morning Tasmeen.        rest is " Tasmeen"    refused
  dictation.   -> dictation.[Image #17]   rest has a space      refused
  Ghost        -> Ghostty                 no space anywhere     kept
  Praisy       -> Praisy's                no space anywhere     kept
  prone.maybe  -> prone point.maybe       not at either end     kept
```

The last row is the one that moved. `Ghost E. -> Ghostty.` also passes: the shared `.` comes off and neither side is a prefix or suffix of the other.

</details>

All 39 cases in `tests/edit-diff-cases.yaml` pass, the five this branch adds and the four #259 added among them. Checked against a line-for-line port of `words`, `trimmed`, `changes` and `added` — there is no Swift toolchain in this session, so CI on macOS is the real run.

---

Issue #262, opened 2026-09-02
Turn the vocabulary gate on for French, refusing 47 of 117 wrong rewrites


znat:
**Depends on:** #261. Until mmBERT-small is in the slot gate there is no multilingual model to turn on.

The vocabulary gate does not run on French dictation. The reason is the model: ModernBERT is English-only, so `SlotReference` has nothing useful to say about a French slot, and `Transcriber.swift` gates the whole stage on `Pipeline.language(of:) == "en"`. With mmBERT-small in place the gate works in French at a floor of about -0.30. Measured on a 201-case French bench: 47 of 117 bad rewrites refused, at a cost of 1 correct rewrite wrongly refused in 84.

The error direction is cheap. `SlotReference` only ever refuses; it cannot write. So its mistake costs one name typed by hand, not corrupted text.

This depends on the mmBERT-small swap landing first.

<details>
<summary>Measured — and the floor is per language</summary>

    on 135 honest cases     AUC     at -0.30: wrong refusals -> caught (of 117)
    ModernBERT              0.677        5 -> 37
    mmBERT-base             0.824        2 -> 49
    mmBERT-small            0.788        1 -> 47

**The floor must be per language.** -0.40 is the only value free in both, and it
costs the English gate 35 of the 55 declines it catches today. Keep -0.20 in
English and use about -0.30 in French. That is the second parameter that keys
off language — the boundary mark set is the other — which suggests one language
block rather than two scattered settings.

**Mechanism.** Both models answer in French; English-only fillers are ~0.1 per
list of 10 for both. The difference is that mmBERT gives **4.79** distinctively
French words per list against ModernBERT's **2.71**, which pads the rest with
`le/la/il/elle/cela`.

`Qwen3-Embedding-0.6B` handles French fine: synonym-minus-unrelated margin
+0.259 against English's +0.407, translation-vs-random 10/10, and accents
survive the tokenizer (`château`/`chateau` 0.914).

Bench and scripts: `~/Documents/parrotflow-scratch/sentence-join/french-slot/`,
full write-up in `FRENCH_SLOT_RESULT.md`.

</details>

<details>
<summary>Why the French number is provisional, and what would settle it</summary>

**The bench is 184 invented cases of 201, with only 2 real declines.** The
user's history holds 878 French lines, but almost every one discusses the
vocabulary problem rather than dictating with the terms in it. Three real ASR
mishearings were salvaged (`Versal`→Vercel, `Prazi`/`Prissy`→Praisy). The
[...]

Issue #260, opened 2026-09-02
Score three readings with Qwen instead of a threshold, to repair 82% of false sentence breaks instead of 26%


znat:
**Depends on:** nothing. **Blocks:** #256 (easier after this), and with #261 it removes ModernBERT from the app.

A pause mid-sentence makes the recogniser write a period and capitalise the next word. The stage that repairs this fixes 26% of them, and it cannot be tuned higher, because its threshold is set by the single lowest-scoring real sentence ending out of 180. Replacing the score with a choice between readings fixes 82% and calibrates nothing. Measured in Swift with the 4-bit model that would ship: 50 ms median per boundary, zero wrong joins on 180 real endings.

Today `SentenceProbe.at()` masks the boundary and computes `log P(".") − log P(" next word")`, and `Config.swift:1674-1675` compares it to `joinBelow: -4` and `offerBelow: -2`. Instead, build three readings of the same text, score each as a whole sequence per token, and take the winner.

<details>
<summary>The calculation, exactly</summary>

For a `word. Capital` boundary, take up to 12 words either side. Build three
readings and score each with a causal model:

    ". Term A is longer"     period, next word capitalised
    "  term A is longer"     joined, next word lowercased
    ", term A is longer"     comma, next word lowercased

Score = sum of `log P(token | everything before it)` over the continuation,
**divided by the token count**. Highest wins. If `joined` wins, remove the
period and lowercase the next word — the repair itself is unchanged.

Three things are load-bearing:

**Per token, not summed.** On summed log-probability the joined reading wins by
being shortest: 97% repairs and 33 wrong joins.

**The comma is an option, not a term in a subtraction.** Scoring
`max(. , ; :) − log P(next)` gives 64% against the shipped shape's 81%, because
a comma is plausible everywhere and lifts the cuts with the real endings. As a
*reading* it is essential: 26% of real endings pick the comma, which is why none
of them picks `join`. Those are the boundaries the user labelled by hand as
"really a comma".

**`;` and `:` never change a decision in English.** Three readings, not five.
Keep the mark set in config — it is language-dependent, and French uses `:`
where English does not.

</details>

<details>
<summary>Measured — 26% to 82%, zero wrong joins</summary>

Bench: 172 real sentence endings and 140 spurious cuts, from the user's own
dictation. `sentence-join/harness/en_real_nf.json`, `en_cuts_nf.json`,
`en_real_labels.json`.

    shipped ModernBERT at -4          26% repaired    0 wrong joins /172
[...]

PR #268, opened 2026-09-03
feat: the slot gate reads mmBERT-small, which also answers in French

### LINKED ISSUE #261: Replace ModernBERT with mmBERT-small in the slot gate, same accuracy, smaller and faster, and multilingual
**Depends on:** nothing. **Blocks:** #262, which cannot happen until this lands. With #260 it removes ModernBERT from the app.

`SlotReference` masks the slot a word sits in, takes ModernBERT's top-10 fillers, and refuses a vocabulary substitution when the heard word fits the centroid better than the term does. ModernBERT is English-only, which is why the whole gate is English-only. `jhu-clsp/mmBERT-small` ties it on the English bench, is smaller and faster than it, and covers 1,800 languages. Swapping it costs nothing in English and is what makes a French gate possible.

This issue is the swap alone, with the gate still English-only. Turning French on is separate.

<details>
<summary>Measured — a tie in English, at a lower cost</summary>

Bench: `~/Documents/parrotflow-scratch/sentence-join/arms/comps.json`, 238 cases,
140 approve and 98 decline, with ModernBERT's fillers stored. Only step 1 of the
pipeline changes; the embedding, the centroid and the floor are held fixed.

    model             at -0.20    AUC     best at any floor
    ModernBERT-base    192/238    0.899        201
    mmBERT-base        193/238    0.873        200
    mmBERT-small       192/238    0.887        202

    model             params   backbone   vocab    fp32 file   CPU     MPS
    ModernBERT-base   149.7 M   110.3 M   39.3 M    571 MB    36.1 ms  17.0 ms
    mmBERT-base       307.8 M   110.3 M  197.5 M   1174 MB    63.6 ms  29.9 ms
    mmBERT-small      140.9 M    42.2 M   98.7 M    537 MB    29.8 ms  18.1 ms

A tie on accuracy, 34 MB smaller and 6 ms faster. `tie_word_embeddings` is true
on all three, so there is no separate output matrix. ModernBERT-base and
mmBERT-base have the *identical* backbone — base is ModernBERT with a 5x
vocabulary, and all of its extra latency is that vocabulary.

At fp16 mmBERT-small is **269 MB with zero decision changes** of 238. bf16
churns far more; skip it.

Full run: `~/Documents/parrotflow-scratch/mmbert-small/RESULT.md`.

</details>

<details>
<summary>Two hypotheses about why, both wrong — do not repeat them</summary>

**"A 256k vocabulary wastes filler slots on case variants like azure/Azure."**
False. Duplicates dropped per top-10: ModernBERT 0.37, mmBERT 0.36. Identical.

**"ModernBERT supplies more proper nouns, and names are what make a name-vs-name
comparison work."** False both ways. Proper nouns per top-10: ModernBERT 1.19,
mmBERT 1.36 — mmBERT supplies *more*. And the premise inverts: of 106 refusals
across both models, 102 came from filler lists with **zero** proper nouns. A
name in the list pulls the centroid toward the unknown term and shrinks the gap.

</details>

<details>
<summary>The weakness this does not fix</summary>

On the 17 declines where the heard word is **itself a name**, both models refuse
**2 of 17**. The `Versailles Castle` case is in the bench three times and
neither refuses it — gaps of -0.08 to -0.15 against a -0.20 floor.

That is the test's real hole, it is encoder-independent, and swapping models
does not touch it. It is the same conclusion the counter-portrait work reached
from the other side: name-versus-name needs a second mechanism, not a better
first one.

</details>

<details>
<summary>Do not confuse the two jobs ModernBERT does</summary>

ModernBERT serves the slot fillers **and** the sentence-boundary probe. mmBERT
loses the boundary probe badly — 59% against 83% — and 74% of that loss is the
`P(".")` term: it does not know where an English sentence ends. Its medians on
real endings are actually *better* (+5.58 vs +5.42); only the tail differs, and
its tail is a cluster rather than one outlier, so forgiving its worst case gains
nothing.

So mmBERT-small must not be given the boundary job. That stage goes to Qwen —
see the boundary issue. If both land, ModernBERT leaves the app entirely.

</details>

znat:
The vocabulary gate's slot stage masks the word a proposal sits on and asks a
masked language model what belongs there. That model was ModernBERT, which is
English-only, and that is why the gate is English-only. This swaps it for
`jhu-clsp/mmBERT-small`, which ties it in English and answers in 1,800
languages. That is the whole point of the swap: it is not cheaper — same cache
size, and 18 ms per pass against 11 — and it buys nothing in English. It is what
makes a French gate possible. The gate stays English-only here; turning French
on is #262, and the floor is now a value keyed by language with only `en` in it.
With #260 merged, this takes the last pipeline stage off ModernBERT.
`SentenceModel`, `SentenceProbe`, `BPETokenizer` and their two CLI commands are
still here and are still fetched — deleting them is a follow-up of its own.

Start the review at `SlotProbe.at`. The conversion had to take `attention_mask`
as a real input, and that is the one thing in this PR that is not a
straight swap — see the first details block. After that, `SlotModel` is a copy
of `SentenceModel` pointed at the new repository, and `SlotTokenizer` reads
mmBERT's Gemma-style `tokenizer.json` through swift-transformers because the
hand-written `BPETokenizer` cannot.

Closes #261.

<details>
<summary>The tie in the issue was measured unpadded, and the app pads — that cost 12 decisions until it was fixed</summary>

The issue's bench feeds the model one sentence at its natural length. The app
packs `[bos] head <mask> tail [eos]` into a fixed 64 and pads. The accepted
ModernBERT conversion builds `attention_mask` as `ones_like(input_ids)` inside
the graph, so that padding is read as text.

ModernBERT does not mind. mmBERT-small does. On the 238-case bench, scoring
each slot's ten most likely words:

| packing | agrees with the unpadded answer | bench at -0.20 | AUC |
|---|---:|---:|---:|
| padding attended, filled with `<pad>` | 0 / 238 | 180 / 238 | 0.842 |
| `attention_mask` a real input, padding masked out | 238 / 238 | 192 / 238 | 0.888 |

It is not subtle. `That was [MASK] suggestion originally.` answers
`my the a no his your our their her another` with the mask and
`is iz in and work ia for first a part` without it.

`znaat/modernbert-coreml` says taking `attention_mask` as an input does not
convert. With `torch 2.7.0 / transformers 4.48.3 / coremltools 9.0` it does, for
ModernBERT-base as well. So `SentenceProbe` attends its padding too. Nothing
reads it any more, so that is left for the PR that deletes it.

</details>

<details>
<summary>Measured — a tie in English, no decision lost to Core ML, and what it costs</summary>

**The vocabulary bench.** `arms/comps.json`, 238 cases, 140 approve and 98
decline. Only step 1 changes: the embedding, the centroid and the -0.20 floor
are held fixed, and every arm is packed the way `SlotProbe.at` packs.

| model | at -0.20 | approve | decline | AUC | best at any floor |
|---|---:|---:|---:|---:|---:|
| ModernBERT-base | 192 / 238 | 136 / 140 | 56 / 98 | 0.887 | 201 |
| **mmBERT-small** | **192 / 238** | 137 / 140 | 55 / 98 | **0.888** | 202 |

Eight decisions differ.

Through the shipped binary — `--slot-gap` on all 238, which uses the 4-bit MLX
Qwen where the table above uses the fp32 one — mmBERT-small scores **191 / 238**
(approve 134/140, decline 57/98). Eleven decisions differ from the Python
column, net -1; the gaps agree to a median of 0.006 and a p90 of 0.044. That is
the embedding, not the slot model.

**`scripts/check-slot-gate.sh`**, the 50 English cases of `tests/judge-cases.yaml`:

| model | applied by the word lists | declined by the slot | left for the judge | errors |
|---|---:|---:|---:|---:|
| ModernBERT-base | 12 | 14 | 24 | 0 |
| mmBERT-small | 12 | 15 | 23 | 0 |

**Core ML against PyTorch**, same 238 cases, compute units pinned to CPU and
GPU: **0 decision flips**, largest gap difference 0.031, top word identical 238/238, same ten
words 230/238, and correlation at the mask 0.99999 to 1.00000 over a 40-case
sample.

**On the Neural Engine the model is wrong**, not approximately right: 0 of 238
filler lists match, the top word matches 18/238, and correlation at the mask
falls to 0.156. So
`SlotModel.load` pins `.cpuAndGPU`. `ALL` picked the same path as `.cpuAndGPU`
on this Mac in all 238 cases, but that is a scheduler decision, not a promise.

**Cost.** The weights are 282 MB against ModernBERT's 299, but mmBERT's
`tokenizer.json` is 18 MB against 2, so the cache is the same size: 286 MB
against 288. A call is 15 ms from Swift (median of ten runs, machine idle) and
18 ms through coremltools, against 11 ms for the same conversion of
ModernBERT-base. mmBERT-small is the smaller model — 42 M of backbone against
110 M — and still slower, because `tie_word_embeddings` makes its 256,000-word
vocabulary the output matrix and every call produces a `[1, 64, 256000]` tensor.
The 15 ms includes reading one row out of that and ranking it, so the readout is
not the problem and tracing a gathered output would not pay. First run
downloads and compiles in 98.6 s; after that the compiled model loads in 0.1 s.

**Other check scripts**, all passing: `check-slot-gap.sh` (decisions held, every
refusal below -0.38 and every no-opinion above +0.02), `check-slot-tokenizer.sh`
21/21, `check-word-gate.sh`, `check-portrait.sh`, `check-counter-portrait.sh`
20/0/0 held out, `check-sentence-open.sh`, `make test`.

</details>

<details>
<summary>The model is published, pinned, and its tokenizer is checked against HuggingFace's own</summary>

`znaat/mmbert-small-coreml`, revision `a983542a6ae292d0147cbdb9b54eddd00c4f6001`.
`SlotModel.revision` and `scripts/check-slot-tokenizer.sh` both pin it, so a new
upload has to be adopted rather than arriving on its own. `HubDownload` grew a
`revision` argument for that; it defaults to `main`, so nothing else changes.

The repository holds two files `znaat/modernbert-coreml` does not — `config.json`
and `tokenizer_config.json`. `AutoTokenizer.from(modelFolder:)` reads all three
and throws without them.

`tests/slot-tokenizer-cases.json` is the new fixture: 21 cases covering the `▁`
word-start marker, byte fallback, accents, the added tokens, and the empty
string, which must encode to nothing rather than to a lone `▁`. It runs in
`make test` and in CI, on the tokenizer files alone.

One documented divergence. `tokenizer_config.json` marks `<mask>` `lstrip`, and
the file's own `added_tokens` does not. `transformers.AutoTokenizer` applies it,
`tokenizers` and swift-transformers do not, so `"it <mask> that"` tokenizes
differently in the two. The app never encodes `<mask>` as text — `SlotProbe.at`
puts the id in — so the fixture describes the path the app takes.
`scripts/slot-probe-reference.py` says so in its header.

</details>

<details>
<summary>What this does not fix, and why SlotProbe is a copy of SentenceProbe</summary>

**Name against name is still the hole.** #261 measured it: on the 17 declines
where the heard word is itself a name, both models refuse 2, and `Versailles
Castle` is in the bench three times with neither refusing it. Nothing here
changes that. It is encoder-independent and a different mechanism has to answer
it.

**One case now returns fewer than three fillers** and the gap comes back 0, which
is "no opinion". It is case 0, where the masked word is the last word of the
sentence. ModernBERT returned four fillers there. Both score it correctly.

**`SlotProbe`, `SlotModel` and `SlotTokenizer` duplicate `SentenceProbe`,
`SentenceModel` and `BPETokenizer` rather than sharing a base.** That is
deliberate and the headers say so: the sentence trio leaves the app with
ModernBERT, so anything the two shared would leave with it. Sharing now would
mean writing an abstraction and deleting it in the next PR.

**ModernBERT is not deleted here.** After this, nothing in the pipeline reads
it, but `--sentence-model`, `--sentence-probe` and `Transcriber.warmSentenceModel`
still do, and it is still downloaded at launch. Taking it out is a small PR of
its own, and it should also drop the 300 MB fetch.

</details>

- [x] `make test` passes.
- [x] A prompt or pattern change is scored, and the numbers are above.
- [x] Every commit is signed off — `git commit -s`. See [CONTRIBUTING.md](../CONTRIBUTING.md).


znat:
Fixed in 9725a0e. `--warm` now reads "the word vectors, the sentence model and the slot model" and says about 1 GB.

Two notes on the same area, both left for a follow-up rather than done here:

- With #260 merged, no pipeline stage reads ModernBERT any more. `--warm` still fetches it, which is 300 MB that buys nothing. Taking it out belongs with deleting `SentenceModel`, `SentenceProbe` and `BPETokenizer`, in one PR rather than scattered.
- `--warm` does not prepare the Qwen sentence readings that #260 added, so a one-shot run still cannot see the boundary stage. Same follow-up.

PR #282, opened 2026-09-04
feat: a term's portrait starts from its first use and its first counter-example

znat:
A term's portrait now starts from one confirmed use and one counter-example. It
needed three of each. The comparison is two centroids — a sentence is written
when it sits closer to the term's own sentences than to the ones it was
corrected out of — and nothing in that comparison needs three. **This is not
measured.** The three was measured on the floor path, before counter-examples
existed, and the comparison was never measured at a low count. It ships on the
user's judgement, to be judged in use.

The floor path keeps three uses. The floor is a leave-one-out quantile over the
uses, so it needs uses to leave out, and the comment on record says two put it
below every ordinary word. So a term with no counter-example and fewer than
three uses now has no verdict at all: `read` returns nil and the stage decides
as it did before.

Start the review at `TermPortrait.summary(for:)`, which is where the new rule
lives, and at `Summary.floor`, which is now optional. `make test` is green and
both portrait scripts were run on this machine — CI does not run them, they
need the 400 MB word-vector model.

This was cut from `issue-264` at 68a018f. That branch was squashed into `main`
as #281 while the work was in progress and deleted, so the commit is
cherry-picked onto `main` and every check was run again on that base.

<details>
<summary>The three numbers, and what each one gates</summary>

| constant | was | is | gates |
|---|---|---|---|
| `TermPortrait.minimum` | 3 | 1 | uses needed for a centre at all |
| `TermPortrait.counterMinimum` | 3 | 1 | counters needed for the comparison |
| `TermPortrait.floorMinimum` | — | 3 | uses needed for the floor |

Which gives:

| uses | counters | read by |
|---|---|---|
| 1 or more | 1 or more | the comparison |
| 3 or more | 0 | the floor |
| 1 or 2 | 0 | nothing — `read` returns nil |
| 0 | any | nothing |

`Summary.floor` and `Reading.floor` are `Double?`. A portrait built from fewer
than three uses has no floor, and `quantileOf` of an empty list is infinity,
which `JSONEncoder` refuses to write — one such term would have failed the
cache write for every term.

</details>

<details>
<summary>Script numbers — both benches are unchanged, one term changed route</summary>

`scripts/check-counter-portrait.sh`, 24 cases over 4 terms, before and after:

| arm | before | after |
|---|---|---|
| floor (counters stripped) — 20 held out | 17 right, 2 wrong, 1 quiet | 17 right, 2 wrong, 1 quiet |
| floor — all 24 | 19 / 4 / 1 | 19 / 4 / 1 |
| comparison — 20 held out | 20 right, 0 wrong, 0 quiet | 20 right, 0 wrong, 0 quiet |
| comparison — all 24 | 23 / 1 / 0 | 23 / 1 / 0 |

Every case gives the same verdict. One term moved: Qwen has 6 uses and 2
counters, so it was below the old `counterMinimum` and read by the floor. It is
now read by the comparison. Its four cases were right before and are right
after. On "Gwen booked the meeting room for Thursday.":

    uses 6   against 2   tightness 0.865   floor 0.882
    score 0.642   counters 0.734   refuses

The sentence scores the same either way. Before, 0.642 was read against the
floor of 0.882 and fell more than `refusal` below it, so it was refused. Now it
is read against the counter centre at 0.734, and refused for that reason.

The other three terms have 4, 5 and 19 counters, so their route is unchanged.
The floor arm strips every counter, so nothing in it could move.

`scripts/check-portrait.sh` passes before and after. Two of its cases had to be
rewritten, because they asserted the old numbers:

- "two counters are too few, so the floor still decides" asserted that the
  Vercel line carried no second score. One counter is now enough, so it is now
  a `check` that it does.
- "the third turns it into a comparison" is now "one counter turns the floor
  into a comparison", and the refusal it used to assert alongside moved down
  past the second and third counter. See the next block for why.

Four cases were added for the new behaviour, which nothing covered: one use and
no counter is not a portrait; one use and one counter is, and its floor column
reads `—`; a sentence like the use is authorised; a sentence like the counter is
refused.

`scripts/check-term-uses.sh` line 237 asserted "a counter does not count toward
the portrait minimum" by showing that two uses and a counter print "no portrait".
Two uses and a counter is now a portrait, so that sentence is no longer true.
It is rewritten to what is still true and still needs no model: one counter and
no use prints `no portrait: 0 confirmed use(s), 1 counter-example(s)`. This
script is in `make test`, so it must never load the 400 MB model.

</details>

<details>
<summary>A one-sentence side is divided by a tightness of 1.0 — this is the thing to watch</summary>

`middle(of:)` over one vector returns that vector and a tightness of 1.0: the
cosine of a vector with itself is 1. Both sides of the comparison are divided by
their own tightness, so a one-sentence side is divided by 1.0 while a
many-sentence side is divided by something below it, which lifts the
many-sentence score. The scale is left exactly as it is; the user chose to try
it that way.

It is visible in `check-portrait.sh`. Vercel with 3 uses and 1 counter, scoring
"Have you ever walked the grounds at Versailles?":

    3 uses, 1 counter    score 0.745   counters 0.662   authorises
    3 uses, 3 counters   score 0.745   counters 0.868   refuses

The verdict is right only once the counter side has more than one sentence. The
comparison is there from the first counter; it is not yet decisive. That case is
now asserted at three counters, and the one-counter step asserts only that the
route changed.

</details>

<details>
<summary>Every stored portrait is rebuilt once</summary>

A portrait is cached under a fingerprint over the term's stored sentences.
Changing a minimum changes what a stored summary means while the sentences stay
the same, so a term with three uses and one counter would have gone on serving
the floor verdict it cached before this change, until the next correction.

`fingerprint(of:)` now carries the three minimums, so every portrait on disk is
stale and is rebuilt once. This also keeps the two bench scripts honest: the
cache lives in Application Support and not in `PARROTFLOW_CONFIG_DIR`, so an
"after" run would otherwise have read the "before" run's summaries.

</details>

<details>
<summary>Docs</summary>

`docs/transcription.md` and `docs/pipelines.md` said three in three places. They
now say one, and say the floor still needs three uses when a term has no
counter-example.

</details>

### DIFF
--- a/Sources/ParrotFlow/TermPortrait.swift
+++ b/Sources/ParrotFlow/TermPortrait.swift
@@ -8,12 +8,13 @@ import Foundation
 ///   - `centre`, the mean of the context vectors — every token of a sentence
 ///     except the term's own, so it describes the sentence and not the word;
 ///   - `tightness`, how close those sentences sit to that centre;
-///   - `floor`, what a genuine use of this term scores against it.
+///   - `floor`, what a genuine use of this term scores against it, from
+///     `floorMinimum` uses up.
 ///
 /// A new sentence is scored against the centre and divided by the tightness.
 /// Above the floor, the rewrite is authorised.
 ///
-/// A term that has been corrected out of enough sentences gets a second centre
+/// A term that has been corrected out of one sentence gets a second centre
 /// built the same way from those. Then there is no floor: the sentence is
 /// written when it is closer to the first centre than to the second, and
 /// refused when it is closer to the second. See `band`.
@@ -37,10 +38,18 @@ actor TermPortrait {
 
     static let shared = TermPortrait()
 
-    /// Fewer than this and the numbers mean nothing: at two uses the floor came
-    /// out below every ordinary word, at three it holds. Under it, a term has
-    /// no portrait and the stage decides as it does today.
-    static let minimum = 3
+    /// How many confirmed uses a term needs before it has a centre at all.
+    ///
+    /// One. A single sentence is enough to compare against a counter-example.
+    static let minimum = 1
+
+    /// How many uses the floor path needs.
+    ///
+    /// The floor is a leave-one-out quantile, so it needs uses to leave out: at
+    /// two the floor came out below every ordinary word, at three it holds. A
+    /// term with no counter-example and fewer than this has no portrait, and
+    /// the stage decides as it does today.
+    static let floorMinimum = 3
 
     /// Where the floor is read off the leave-one-out scores.
     static let quantile = 0.10
@@ -141,11 +150,10 @@ actor TermPortrait {
     /// How many counter-examples a term needs before the comparison replaces
     /// the floor.
     ///
-    /// Below this the counter centre is untestable, and the one term measured
-    /// with two counters got both of its own cases wrong. 3 costs nothing on
-    /// either bench set and is `minimum` again, so there is one number here and
-    /// not two.
-    static let counterMinimum = 3
+    /// One. The three that stood here was measured on the floor path, before
+    /// counter-examples existed; the comparison was never measured at a low
+    /// count.
+    static let counterMinimum = 1
 
     /// How far apart the two scores have to be before either side wins.
     ///
@@ -158,7 +166,8 @@ actor TermPortrait {
     struct Summary: Codable, Equatable {
         let centre: [Float]
         let tightness: Double
-        let floor: Double
+        /// nil below `floorMinimum` uses, where there is nothing to leave out.
+        let floor: Double?
         /// The sentences this was built from, so a portrait is recomputed when
         /// they change and not otherwise.
         let fingerprint: String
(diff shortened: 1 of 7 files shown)


znat:
Greptile's runner has no Swift toolchain, so it could not build or run anything. Here is that verification, done on the macOS machine, on 48aa35e — the commit it reviewed.

```
swift build -c release                  Build complete!
swiftlint lint --quiet                  no warning in either changed Swift file
make test                               Every check passed. (35 check scripts)
scripts/check-portrait.sh               Every check passed. (14 checks)
scripts/check-counter-portrait.sh       Every check passed.
```

`scripts/check-term-uses.sh` is one of the 35 in `make test`, so it ran there.

`check-counter-portrait.sh` needs the 400 MB word-vector model, which is why it is not in `make test`. Its numbers are identical before and after this change, in both arms:

```
floor arm        20 held out: 17 right  2 wrong  1 quiet     all 24: 19 / 4 / 1
comparison arm   20 held out: 20 right  0 wrong  0 quiet     all 24: 23 / 1 / 0
```

One term changed route. Qwen has 6 uses and 2 counters, so it was below the old `counterMinimum` of 3 and read by the floor; it is now read by the comparison. All four of its cases give the same verdict as before.

PR #284, opened 2026-09-05
feat: a priority list of microphones, set from the menu bar

znat:
`audio.microphones` is a priority list. The first microphone in it that is plugged in is the one ParrotFlow records through, whatever System Settings has the default input set to. Only this app moves — it opens the device for itself, so Zoom keeps using what it was using. The list is set from the menu bar: the `Microphone · …` row now unfolds, and each entry can be moved up, moved down, removed, or made the one recording now.

Empty is the old behaviour, byte for byte: no list, no pin, follow the system. But a config that has never carried the setting gets a list written the first time the menu is opened — everything attached, with the mic already in use at the top. That changes no microphone; it writes down the one you were already on, so System Settings moving the default stops moving this app.

Reviewers: start at `Recorder.desiredInput()`. It is the one place that decides both what the engine is bound to and which device to open, and everything else in `Recorder.swift` is the existing change-detection now asking it instead of asking the system default.

<details>
<summary>The pin was measured, not assumed — the tap format follows the pinned device, not the default</summary>

The load-bearing question was whether an `AVAudioEngine` can be pinned to a non-default input without breaking the format checks `Recorder` already does. It can, with `inputNode.auAudioUnit.setDeviceID`, and it has to happen before any format is read off the node.

Measured on this machine, with the built-in mic at 44100 Hz as the system default and a second device at 48000 Hz pinned:

| | sample rate |
|---|---|
| system default input | 44100 Hz |
| pinned device | 48000 Hz |
| node tap format | 48000 Hz |
| node hardware format | 48000 Hz |

Both halves of the node follow the pin. So `graphProblem` and `formatProblem` keep working unchanged — as long as `formatProblem` compares against the device we pinned to rather than the default, which is what `desiredInput()` is for.

The discriminating end-to-end test used the Teams loopback driver, which is silent:

```
microphones: teams   → through Microsoft Teams Audio, peak 0.000  (silent)
microphones: (empty) → through MacBook Pro Microphone, peak 0.33
```

The system default was the built-in mic in both runs.

</details>

<details>
<summary>Why the device list is watched as well as the engine</summary>

`AVAudioEngineConfigurationChange` is the engine talking about its own device. An engine pinned to one microphone has nothing to say when a better one is plugged in, so a priority list would only ever be read at launch and on a save of `config.yaml`.

`Recorder.watchDeviceList()` adds a listener on `kAudioHardwarePropertyDevices`. It is ignored while recording: a device arriving is not a reason to take the microphone away from a sentence somebody is halfway through, and the next press picks it up.

It is also ignored until something has bound an engine. Instantiating an input node is what makes AVFoundation put up the native microphone dialog, and at launch that would come before the permissions window has explained why — the same reason `warmUp` is already gated on the permission. `reevaluateInput()` returns while `bound` is nil.

</details>

<details>
<summary>Two things a config file can say, and only one of them seeds</summary>

The seed has to happen once. Without a way to tell "never decided" from "decided to follow the system", `Follow the System Setting` would clear the list and the next menu open would refill it.

- key absent — nobody has been asked. The menu seeds.
- `microphones: []` — somebody said follow the system. Left alone.

Both decode to an empty array, so the difference is asked of the file text, in `ConfigWriter.hasMicrophonesKey()`.

</details>

<details>
<summary>Matching an entry to a device, and what it does when two answer to the same name</summary>

An entry is a name, CoreAudio's UID, or a fragment of a name — `AirPods` reaches `Nathan's AirPods Pro`. Case is ignored. An exact match over every attached device wins before any fragment is tried, so a device called `Display Audio` does not lose to one called `Display Audio (2)`.

Two microphones can answer to the same name, and that is what the UID is for. `--microphones` prints both. The seed writes the UID instead of the name when the name is ambiguous.

</details>

<details>
<summary>Three CLI doors, because the list is typed by hand as often as it is clicked</summary>

```sh
$PF --microphones                          # what is attached, with names and UIDs
$PF --microphones --set "Studio Mic, Mac"  # write the order, best first
$PF --microphones --set ""                 # follow the system
$PF --check-config                         # which entries are attached, which wins
$PF --record 2                             # says which microphone it recorded through
```

`--microphones` is where the strings in the setting come from. A name typed from memory is a name that matches nothing.

</details>

<details>
<summary>What is not verified here — hot-plug, on a Mac with nothing to unplug</summary>

This machine has one real microphone and two virtual driver devices. Nobody has watched the list fail over to entry 2 when entry 1 is unplugged, and back again. The listener is wired and the rate-change test showed device IDs do move under CoreAudio, but the failover itself wants a headset and a pair of hands.

Also worth knowing: CoreAudio calls a virtual driver an input device, so `ZoomAudioDevice` and the Teams loopback are seeded into the list like anything else. They are exactly the wrong thing to fall back to — the clip comes back silent. They are removed from the menu in one click, and the docs say so.

</details>

<details>
<summary>Tests — scripts/check-microphones.sh, 25 cases over the four shapes config.yaml can be in</summary>

The menu writes into a file that is mostly comments, so the write is a text splice rather than a YAML round trip. Every shape the file can be in has to come out valid with its comments intact:

- an `audio:` block with no list yet
- a list already there, replaced
- a list a person wrote as a flow sequence, `microphones: [a, b]`
- no `audio:` block at all
- emptied to `[]`, then filled again
- a name that is not a plain word, `Nathan's AirPods: Pro`

25/25. Wired into `.github/workflows/checks.yml`. `check-audio-recovery.sh` 14/14, `check-learn.sh` 45/45 and `check-default-config.sh` still pass.

</details>

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -234,6 +234,12 @@ jobs:
       - name: --learn writes a pronunciation
         run: scripts/check-learn.sh
 
+      # The menu bar writes the microphone priority list into config.yaml, a
+      # file that is mostly comments — so the write is a text splice, and every
+      # shape the file can be in has to come out valid and keep its comments.
+      - name: The microphone list survives a round trip
+        run: scripts/check-microphones.sh
+
       # Two ways in, one signing identity. A check that is only right in the
       # script or only right in the app is a door that checks less than the
       # other one.
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -5871,10 +5871,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 // it below one that is already plugged in would change nothing
                 // and look broken.
                 let item = NSMenuItem(
-                    title: "\(device.name)", action: #selector(useMicrophone), keyEquivalent: ""
+                    title: device.name, action: #selector(useMicrophone), keyEquivalent: ""
                 )
                 item.target = self
-                item.representedObject = device.name
+                item.representedObject = Self.entry(for: device, among: attached)
                 item.state = device.uid == live ? .on : .off
                 menu.addItem(item)
             }
@@ -5929,15 +5929,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // and `sorted` given a predicate that is not one is allowed to return
         // anything at all.
         let ordered = attached.filter { $0.uid == first } + attached.filter { $0.uid != first }
-        // The name, unless two microphones answer to it — then the UID, which
-        // is the thing that tells them apart. Written once, and read by
-        // everyone afterwards.
-        let names = attached.map(\.name)
-        let seeded = ordered.map { device in
-            names.filter { $0 == device.name }.count > 1 ? device.uid : device.name
-        }
+        let seeded = ordered.map { Self.entry(for: $0, among: attached) }
         writeMicrophones(seeded)
         return seeded
+    }
+
+    /// What to write in the config for one microphone.
+    ///
+    /// The name, unless a second attached device answers to it — then the UID.
+    /// Two microphones can share a name, and a name is then not enough to pick
+    /// the one somebody clicked: `Recorder.device(named:)` would resolve it to
+    /// whichever CoreAudio lists first.
+    private static func entry(
+        for device: Recorder.InputDevice, among attached: [Recorder.InputDevice]
+    ) -> String {
+        attached.filter { $0.name == device.name }.count > 1 ? device.uid : device.name
     }
 
     /// What one entry in the list can be moved to.
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -894,6 +894,13 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // on any save of config.yaml.
         Trace.directory = config.resolvedOutputDir
 
+        // Re-read on every save, and acted on: a microphone added to the list
+        // is one the next press should already be listening through, not one
+        // that waits for a relaunch. `reevaluateInput` rebinds only when the
+        // list now names a different device.
+        recorder.preferredMicrophones = config.audio.microphones
+        recorder.reevaluateInput()
+
         configProblems = config.problems()
         for problem in configProblems { Log.write("config: \(problem)") }
         // Logged and not flashed. A `command:` transform is announced on every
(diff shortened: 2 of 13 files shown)


znat:
All four fixed in 56e05db.

**Handle failed device pinning** — right, and worse than described: nothing could ever correct it. `reacquireIfInputMoved` compares `desiredInput().binding` against `bound`, and both named the preferred device, so the check agreed with itself forever while the words came from the system default. Two changes. `pin` now returns whether the device opened, and the caller records what the engine is really on. And the device that refused goes into `unopenable`, so the resolution falls past it to the next entry — a microphone that will not open is the same to this list as one that is unplugged. The set is cleared whenever the device list changes, because the IDs move with it and a stale entry would skip an innocent device.

**Preserve microphone identity** — correct, and it was inconsistent with the seeding path in the same file. Both now go through `AppDelegate.entry(for:among:)`, which writes the UID when a second attached device answers to the same name.

**Replace the complete list** — correct. `endOfList` now consumes comments and blank lines that sit between entries, and stops at the last `- ` item, so a comment after the list stays with whatever follows it. Case added: `check-microphones.sh` writes a list with `# the one in the meeting room` between two entries and checks both entries go.

**Reject a missing value** — fixed, exit 1 with `--set needs a value`. `--set ""` still clears. Case added.

`check-microphones.sh` is 30/30. `check-audio-recovery.sh` 14/14, `check-learn.sh` 45/45, `check-default-config.sh` passes.

One note for anyone reading the artifacts: a second real microphone arrived on the test machine while these were being fixed, and pinning past the system default was measured again with two real devices — a USB mic as the system default, the built-in pinned. It records through the pinned one. The built-in returns digital silence on that machine at the moment, but that is the hardware, not this change: `AVCaptureSession` reads the same `0.00000` from it without going anywhere near `setDeviceID`.

znat:
Fixed in ae53725.

**Retry unavailable microphones** — correct. A device can refuse for a reason that passes, another app holding it or a link still settling, and nothing announces it coming back. A refusal now carries the time it happened and expires after 60s, checked in `desiredInput()` where the resolution reads it. No timer: nothing fires on its own, so the retry happens on the next press or config reload after the minute is up, which is the only moment it is worth anything. A device-list change still clears the whole set early.

60s is the trade. Shorter retries a broken microphone on nearly every press; longer leaves you on the wrong one after it recovers.

**The hot-plug case is no longer untested.** The PR body says nobody had watched the list fail over, because this machine had one real microphone. A USB mic was plugged in and then unplugged while these fixes were being made, and both halves happened:

```
RØDE attached, list "RØDE, MacBook Pro Microphone"  → through RØDE VideoMic GO II
RØDE unplugged, same list                           → through MacBook Pro Microphone, peak 0.30
```

The second is entry 1 gone and entry 2 taken, with no restart.

It also explains the silent built-in mic mentioned in the last comment: it read `0.00000` only while the RØDE was attached, and it is back at 0.30 now that the RØDE is gone. Not this change, and not the hardware either.

check-microphones.sh 30/30, check-audio-recovery.sh 14/14.

PR #288, opened 2026-09-06
fix: a press that did nothing now works, or says why

znat:
Pressing the hotkey did nothing about half the time, and starting a dictation could refuse twice in a row with "The microphone changed as the recording started — press again". Idle, the app rebuilt its capture engine every ten seconds, forever: 40 rebuilds between 19:02 and 19:17 on 2026-09-06, each one opening the microphone. Three separate causes, each found by putting a log line where there had been a silent `return`. A fourth failure — Escape and the offer's letters going dead — turned out to be another app holding Secure Event Input, which nothing can work around, so the app now names that app and says what to do about it.

Start the review at `Recorder.captureFormat` and `ModifierKey.sawInputBeforeTheMonitors`. Those two are the whole of what a person feels. `KeyboardNotice.swift` is 276 of the 673 added lines and changes no existing behaviour.

`--audio-recovery` goes from 14 checks to 16, and the four under `Capture after a device change` pass again after failing silently for weeks.

<details>
<summary>The microphone refused presses that were going to work, and the refusal ran the app in a circle</summary>

The input node holds two formats for itself: the tap format and its own hardware format. They can disagree for as long as the machine stays as it is — 24000 Hz and 48000 Hz on an idle Mac with AirPods on it. `graphProblem` called that unusable, so a press was refused before anything was tried, and the press after it was refused the same way.

That pair is only ever compared by `installTap`, and its exception names which half it asserts on:

```
required condition is false: format.sampleRate == inputHWFormat.sampleRate
```

So the tap is now installed at the hardware format, which cannot raise, and the check in front of it is gone — it could only ever refuse a press that was going to work. `captureFormat` hands back the tap format whenever the two rates agree, so nothing moves on a healthy machine.

The loop was the same mismatch seen over and over. A rebuild retires the engine it replaced; releasing that engine ten seconds later posts another configuration change; the change found the mismatch and rebuilt again. A mismatch that outlives its own rebuild is now remembered, and every rebuild sets that fresh, so a device that moves and then settles gets its chance again.

Confirmed on a real device change at 21:36 — the RØDE was replaced by the built-in microphone and the two formats came back 44100 and 48000. One rebuild, then it recorded. Under the old code that press was a refusal.

Two new `--audio-recovery` cases cover it: one mismatch buys one engine and not a hundred, and the tap goes on at the hardware format.

</details>

<details>
<summary>The hotkey worked one press in two because a system clock stops ticking under secure input</summary>

A press is held back for `press_delay_seconds` and delivered only if nothing else happened in the window. The cheap half of that test compares two ages from `CGEventSource.secondsSinceLastEventType`: the modifier's own down edge, and the last key or click. Anything newer than the modifier arrived after it, which is what a shortcut looks like.

Secure Event Input stops the session source recording the keyboard, and the reading freezes. The poll saw the key go down 25 ms ago and the API said the last `flagsChanged` was seconds ago, which makes every click and keystroke inside those seconds newer than the modifier. Whether a press survived came down to whether you had touched anything recently.

Measured on 2026-09-06 while another app held secure input:

```
key: neither press nor tap — a click 1121 ms old against a modifier 4621 ms old
key: neither press nor tap — a key 2734 ms old against a modifier 4234 ms old
key: neither press nor tap — a key 1183 ms old against a modifier 2183 ms old
```

A stale edge abstains now. Past a quarter of a second the reading is not about this press, and the monitors installed a line above are what is left to catch a shortcut. `physicalEdge` already guarded the same API the same way.

The same swallowed press was also why tapping the hotkey stopped opening the offer panel: a spent hold is neither a press nor a tap.

</details>

<details>
<summary>Another app can take your keyboard, and until now nothing said which one</summary>

Secure Event Input is the OS promising that no other process sees the keyboard. Escape stops cancelling a dictation and the offer's letters stop running its commands — and pressing one does two wrong things at once, because the command does not run and the letter goes into the document.

No permission gets past it, and none should. What the app can do is name the app holding it, which is the one thing a person needs and cannot get without `ioreg`. On 2026-09-06 Notion held it from some time between 18:59 and 19:19 until it was quit.

So there is a notice, in the same panel and the same corner as the Bluetooth microphone notice, said once per episode:

> **Notion has taken the keyboard**
> Quit it, or click out of its password field.

`Why` opens onto what stops working, why it happens, and the three ways to clear it: click into an ordinary field in that app, switch away and back, or quit it.

The name costs two reads and no permission — `NSRunningApplication` for an app, `proc_pidpath` for a process that is not one. macOS attributes the flag to the responsible app, so a CLI tool that enables it from a terminal is reported as the terminal.

Dictation is untouched by any of this and keeps working: the hotkey reads modifier flags rather than events.

`--panels keyboard` raises the notice, and both its states are on `--panel-sheet`.

</details>

<details>
<summary>The pill had two forms and gave the failure messages the wrong one</summary>

A notice raised during a dictation hung off the line you were typing on. The same message raised by a press that failed came out as a floating black lozenge with the plumage rim turning on it, in the middle of the screen — because the surface picked its form by asking whether a dictation was on screen, and a press that fails never starts one.

There is one form now. The anchor decides where the surface goes and no longer decides what it is: hanging off the line when there is one, and `Dock.free` when there is not, which is the same tint and the same margin with all four corners rounded. Net −17 lines.

</details>

<details>
<summary>Two commits guessed at a cause, and are reverted here rather than dropped</summary>

`pill.isOpen` reads a state nothing clears on the way out, so a panel taken off the screen goes on saying it is open. That is real, and it was never shown to be what anybody reported. The tap that did not open the panel and the hold taken as an edit instruction have one cause between them, and it is the frozen event clock above.

A change nothing scored does not stay, so both are reverted. The log lines they added stayed — those are what found the real cause an hour later.

Every silent exit on those paths now says why it took it: `summon:`, `key:`, `hold: an edit instruction — …`, and four lines in `EditWatch` that say what it is holding and what it had nothing to hand over.

</details>

<details>

Two reports from the same session are still open. Neither has reproduced since these fixes went in, and both now write a line naming the reason instead of nothing:

- A hold after the offer panel closes being taken as an edit instruction.
- A word corrected by hand that never reached the vocabulary panel. The evidence points at the app being restarted between the correction being held and the Return that would have handed it over, but that is not proven.

`KeyboardNoticeMetrics.height(expanded:)` is a fixed 300pt, the way `MicNoticeMetrics` is. The collapsed state has been seen; the expanded one has not.

</details>

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -1258,6 +1258,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // and the behaviour cannot come apart again.
         if !recorder.isRecording {
             keyedAtPress = afterTap || (offerIsUp && pill.isOpen)
+            // Only when it is on, and it says which of the two put it there.
+            // This is the one decision at the press you cannot see from
+            // outside: the same key, the same meter, and the words routed
+            // instead of written down. A hold taken as an edit when you meant
+            // to dictate left nothing in the log to name the reason.
+            if keyedAtPress {
+                Log.write(
+                    "hold: an edit instruction — \(afterTap ? "the tap before it" : "the open panel")"
+                )
+            }
         }
         // Read before anything this press does, so an abort later can tell the
         // transcription this press started from one that was already running.
@@ -3656,7 +3666,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // stops a recording, and a stop that came out short is a stop — the
         // press was simply never delivered. Summoning there would answer a key
         // meant for the microphone, and do it over the *previous* sentence.
-        guard !recorder.isRecording, runsInFlight <= 0 else { return }
+        // Said out loud, all three of them. Every way out of here used to be a
+        // bare `return`, so a tap that did nothing left a log with nothing in
+        // it between the last dictation and the next one — and "nothing
+        // happens" is exactly how this is reported.
+        guard !recorder.isRecording, runsInFlight <= 0 else {
+            Log.write(
+                "summon: not now — recording \(recorder.isRecording),"
+                + " \(runsInFlight) run(s) in flight"
+            )
+            return
+        }
         // A tap while the offer is already open is a tap at nothing. Raising a
         // second one over the first would take the letters again and restart a
         // clock the pointer may be deliberately holding.
@@ -3664,10 +3684,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // Closed, it is the opposite gesture rather than none: the tab is on
         // screen so that a tap can open it, and the key it draws is this one.
         if offerIsUp {
-            if !pill.isOpen { openTheOffer() }
+            if !pill.isOpen {
+                openTheOffer()
+            } else {
+                Log.write("summon: the panel is already open")
+            }
             return
         }
-        guard config.feedback.correctOffer else { return }
+        guard config.feedback.correctOffer else {
+            Log.write("summon: the offer is switched off in config.yaml")
+            return
+        }
 
         // The selection wins, and it never goes stale: it is what you are
         // pointing at now. It is also the only target this can have in an app
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -99,6 +99,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private let previewPanel = PreviewPanel()
     /// Says once per microphone that this one will cost you words.
     private let micNotice = MicNotice()
+    private let keyboardNotice = KeyboardNotice()
     /// The release notes, and the three answers to them.
     private let updatePanel = UpdatePanel()
     private var pendingSelection: SelectionReader.Selection?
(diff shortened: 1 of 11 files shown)


znat:
Both Greptile findings looked at. One fixed, one rejected on the merits.

**Issue 2, unnamed holder hides notice — real, fixed in ff1b987.**

`SecureInput.holder()` answers `Holder(pid: nil, name: nil)` when the flag is on and the session names no owner, which is a screen lock or loginwindow. `told` is also nil until something is said, so `told != holder.pid` compared nil to nil, found them equal, and returned before drawing. The first episode of a dead keyboard with nobody to name read as one already announced — and that is the episode where a person has no other way to find out what happened. `told` is an episode now, `.app(pid)` or `.unnamed`, with nil kept for "nothing said yet".

**Issue 1, secure input opens the microphone — the observation is right, the fix is not.**

The behaviour change is real: under Secure Event Input a modifier followed by an arrow key can now reach the delayed dictation instead of being read as a shortcut. Secure input blocks the keyboard for every other process, so the `NSEvent` monitors in `watchForOtherInput` do not see the arrow either. This PR's own third section is the evidence — Escape stops cancelling a dictation and the offer's letters stop running its commands for exactly that reason. There is no path, permitted or otherwise, that observes a key while another app holds the flag.

The proposed fix is to suppress the hold when no reliable shortcut-observation path is available. Under secure input there never is one, so that reads as: the hotkey does nothing at all while another app holds the flag. On 2026-09-06 Notion held it from somewhere between 18:59 and 19:19 until it was quit — twenty minutes with no dictation, where today there is roughly half of it and after this PR there is all of it. That is the bug this PR was opened to fix, put back and made total.

So the trade is deliberate: dictation always works, and a modifier-plus-arrow shortcut typed during someone else's password field can open the microphone. Losing a shortcut is visible and recoverable — release the key, and a clip under `min_duration_seconds` is thrown away. A hotkey that silently does nothing for twenty minutes is neither.

Mouse events are not blocked by secure input, so clicks are still caught by the monitors. The residual gap is keyboard shortcuts during someone else's secure-input episode, and nothing in the process can close it.

Leaving `sawInputBeforeTheMonitors` as it stands. Happy to be overruled — it is a product call about which failure is worse, not a defect.

---

PR #290, opened 2026-09-07
fix: a name is not written over the spelling it was corrected from

znat:
Correcting one spelling of a name, in a sentence that also holds another, taught the name from the wrong half. "I work with Jimmy and my friend is Jimmie." was stored whole as a use of `Jimmie`, so its portrait learned that a correct `Jimmy` nearby is evidence for writing `Jimmie`. On a seeded portrait that single row moved two of four decisions, both toward writing the name over a word that was right.

A stored sentence is now cut at the nearest other spelling on each side, on both the side that builds a portrait and the side that scores a new sentence. There is no setting and no threshold: the spellings to cut at are the ones the user actually corrected.

Start review at `TermPortrait.rivals(of:in:)` — if that set is wrong, everything downstream is. `scripts/check-counter-portrait.sh` is unchanged at 23/24 overall and 20/20 with no error on the held-out set.

<details>
<summary>Why the old behaviour was wrong — the stored side was never windowed</summary>

`SentenceGate.swift:106` windows the sentence being scored to five words either side of the span. `TermPortrait.build` did no such thing: it passed the whole stored `use.said` to `WordVectors`. So the two sides were measured differently, and a stored sentence carrying the other spelling put it straight into the term's centre.

This PR does not add the radius window to the stored side. It cuts at rivals only, so a row with no other spelling in it produces the same vector as before. That keeps the bench result below meaningful — nothing that worked changed.

</details>

<details>
<summary>The cut list is exact, so there is no distance and no threshold</summary>

Three sources, all recorded by the user:

- the term itself
- the `span` of every row the term has — a counter's span **is** the ordinary word, so counters carry their rivals already
- `heard` on a use, the spelling that correction replaced

The third is new, and it is what makes the cut work on a term's **first** correction. Before it, the rival was written to `vocabulary.yaml` as a pronunciation and `TermPortrait` has no Config access, so the spelling was recorded exactly and reachable nowhere the cut could see.

Two details worth checking:

- Matching ignores case. A spelling that opens a sentence is capitalised there and not in the row that recorded it. Over-cutting costs a few words of context; under-cutting puts the other spelling back in the portrait.
- A rival that overlaps the span is skipped. Otherwise `That was Praisy's suggestion` cuts itself, because `\bPraisy\b` matches inside `Praisy's`.
- `heard` is stored only when it differs from the span. `vercel` corrected to `Vercel` is a capital being fixed, and taking it as a rival would cut the sentence at the term.

</details>

<details>
<summary>Measurements — two of four decisions recovered, bench unchanged</summary>

Seeded `Jimmie` portrait, three uses and two counters, only the first row differing:

| sentence | contaminated row | after the cut | clean row |
|---|---|---|---|
| I work with Jimmy on the parser. | authorises | no opinion | no opinion |
| Jimmy brought the cake to the office party. | no opinion | refuses | refuses |
| My colleague Jimmy reviewed the PR. | authorises | authorises | authorises |

The cut row lands within 0.007 of a genuinely clean one, so keeping the row costs nothing against discarding it.

`scripts/check-counter-portrait.sh`, before and after:

```
the comparison — the counters kept
  the 20 held-out cases   20 right   0 wrong   0 quiet
  all 24                  23 right   1 wrong   0 quiet
```

</details>

<details>
<summary>What this does not fix — the cut leaves the rival's own predicate</summary>

The cut removes the other spelling, not the clause it belongs to. On "Je joue avec Eric au basket et avec Erik au tennis." stored as a use of `Erik`, cutting at `Eric` leaves `au basket` attached:

```
                                before   after the cut   ideal clause cut
Je joue avec Eric au basket.    0.959    0.917           0.898
```

All three authorise. The score falls and the wrong write stands. Even a clause-perfect cut removes only one of the two wrong writes on that row — the rest needs a counter, which is what the floor rule has always needed.

Also unfixed: `TermUses.occurrence` returns the **first** whole-word match, so a correction made on the second `Praisy` of `Praisy asked Praisy's team` still builds its window around the first. `Use.span`'s own docstring says the field exists to prevent this, and a string cannot. Storing the occurrence ordinal would fix it.

</details>

<details>
<summary>Review notes — two traps the change had to carry</summary>

**The cache would have hidden all of this.** A portrait is cached under `fingerprint(of:)`, which hashes the rows and not the code. A pure code change serves stale portraits forever. `rule` now carries `/cut1`, and `heard` joins the hashed row.

**`Summary.uses` and `counters` described what was on disk**, not what was built. Wrong the moment a row is skipped.

A row the cut empties is logged and skipped rather than left to throw. That `try` sits inside the build loop, so a throw there costs the term its whole portrait instead of the one row — the same silent-disappearance shape as the `said.contains(span)` filter in `TermUses.read`.

The three `AppDelegate`/`LearnCommand` call sites that now pass `heard` are the ones that already held the replaced spelling and dropped it. The counter call beside them needs nothing: a counter's span is already the other word.

</details>

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -4064,7 +4064,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             // two things at once, and both are worth keeping: the first does
             // not live here and the second does.
             if let right = existingTerm(named: back) {
-                try TermUses.record(term: right, said: sentence, span: back)
+                // `written` is the spelling this replaced, which is what lets
+                // the portrait cut a sentence holding both of them.
+                try TermUses.record(
+                    term: right, said: sentence, span: back, heard: written
+                )
                 rebuildPortrait(for: right)
             }
             Log.write(
@@ -5178,7 +5182,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                     // The span stays as typed. It has to be the word standing
                     // in the sentence, and the sentence holds what was written.
                     let term = existingTerm(named: rule.corrected) ?? rule.corrected
-                    try TermUses.record(term: term, said: corrected, span: rule.corrected)
+                    try TermUses.record(
+                        term: term, said: corrected, span: rule.corrected,
+                        heard: rule.heard
+                    )
                     rebuildPortrait(for: term)
                 } catch {
                     // A portrait that missed one sentence is worth less than a
--- a/Sources/ParrotFlow/LearnCommand.swift
+++ b/Sources/ParrotFlow/LearnCommand.swift
@@ -30,7 +30,8 @@ enum LearnCommand {
                 // built out of sentences somebody invented behaves differently
                 // from one built out of real dictation.
                 try TermUses.record(
-                    term: corrected, said: sentence, span: corrected, from: .seeded
+                    term: corrected, said: sentence, span: corrected, from: .seeded,
+                    heard: heard
                 )
             } catch {
                 print("! the sentence was not recorded in"
--- a/Sources/ParrotFlow/LearnCommand.swift
+++ b/Sources/ParrotFlow/LearnCommand.swift
@@ -71,7 +71,8 @@ enum LearnCommand {
                 }
                 if said != use.said { cut += 1 }
                 let now = TermUses.Use(
-                    said: said, span: use.span, from: use.from, counter: use.counter
+                    said: said, span: use.span, from: use.from, counter: use.counter,
+                    heard: use.heard
                 )
                 // The custom `==` is on the sentence and the span, so two rows
                 // that narrow to the same sentence collapse into one. The last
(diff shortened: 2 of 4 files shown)


znat:
All three fixed in a34604b8. Both blocking findings were real, and the first is worse than the bug this PR set out to fix.

**`--tidy-uses` discarded every `heard`.** The rebuild at `LearnCommand.swift:73` named four fields of five. One run over an existing file would strip every replaced spelling, silently, and the cut would then have nothing to work from. That destroys evidence already collected rather than merely failing to use it.

**A second correction could never fill in a missing `heard`.** `record` returns early when a row with the same sentence, span and polarity is already stored. That is right when nothing changed — but a row written before this field existed never changes, so it could never gain one. Not a rare path: it is every row that already exists. It now fills in a missing `heard` and writes, and still returns without touching the file when there is nothing new.

**`Summary.counters`** now reports the counters that reached the centre rather than the rows on disk. Its docstring described the behaviour from before this branch changed it and was corrected too.

<details>
<summary>Verified end to end on a scratch config, not just compiled</summary>

```
--tidy-uses          heard: "Eric" before  ->  heard: "Eric" after
second correction    row with no heard     ->  gains heard: "Eric"
```

`scripts/check-counter-portrait.sh` unchanged: 23/24 overall, 20/20 held out, 0 wrong, 0 quiet.

</details>

PR #291, opened 2026-09-07
refactor: the judge is gone, so nothing says it is still there

znat:
No model has been asked about a vocabulary proposal for some time. There is no call in the stage, `SentenceGate` says so in its own comment — *"there is no model, and handing a place back with nobody to hand it to is a keep"* — and a day of the dev log carries no line from it. The code went on describing one anyway, in two places a person actually reads.

This renames what was misnamed and corrects what was untrue. No behaviour changes: `check-word-gate` is 35/35 and `check-slot-gate` is applied 12, declined 15, open 23 with no error of either kind, which is the documented baseline.

Start review at the two user-visible strings below — they are the only part that was wrong rather than merely stale. Everything else is a rename.

<details open>
<summary>The app told you a judge existed — twice, where you could see it</summary>

The setup screen's row for the slot model said the cost of it failing to arrive was:

```
costOfFailure: "the vocabulary gate asks the judge instead"
```

and the log said the same until the download landed. Neither was true. With no slot model nothing reads the slot, so a proposal the word lists cannot settle simply keeps what was heard. Both now say that.

</details>

<details>
<summary>Two renames, and why these names and not others</summary>

**`VocabularyJudge` → `VocabularyPass`** (39 uses, 12 files). It never judged anything. It holds `Caps`, `Standing`, `Slot`, `Change`, the three part-finders, `settle` and `settling` — the matching engine. The docs already called it *the vocabulary pass*, 22 times against 3 for *the vocabulary judge*.

**`SlotGate.Route.judge` → `.open`.** It means "nothing settled this, leave it open", which is what `Pipeline` already logs. This was the worst name left: `route: judge`, printed by `--word-gate`, sends a reader hunting for a stage that is not there.

That rawValue escapes into test data, so `scripts/check-slot-gate.sh`, `scripts/check-word-gate.sh` and 35 lines of `tests/word-gate-cases.yaml` move with it. Five `Log.write` lines that opened `"vocabulary judge:"` now open `"vocabulary pass:"`.

</details>

<details>
<summary>What was deliberately kept, so nobody re-files it</summary>

**`tests/judge-cases.yaml` keeps its name.** Three checks that still run read it — `check-slot-gate.sh`, `check-suggest.sh`, `check-spells-rule.sh`. The labels record what a person wanted, not what read them.

**The judge-tuning scripts and `tests/judge-menus.json` stay.** `scripts/gap-signal.py` reads that cache and is explicitly not judge tuning — *"no model call anywhere in this file"* — so deleting the cache would take a working script with it. They are also the record of how those decisions were made, which is not the same thing as a judge.

**Three notices in `Config.swift` keep the word**, because they tell someone with an old config what their old key was called.

</details>

<details>
<summary>How it was checked</summary>

```
swift build -c release        clean
make test                     green
check-word-gate               35/35
check-slot-gate               applied 12  declined 15  open 23  0 wrong either way
```

The audit that drove the sweep: every present-tense claim that a judge runs, in `Sources/` and the check scripts, is gone. What remains matches `judgement`, `judged`, the fixture filename, and the three migration notices above.

</details>

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -841,7 +841,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// means "not this time" and nothing is remembered.
     ///
     /// Only for what somebody asked for out loud. A `prompt:` stage and the
-    /// vocabulary judge reach the same models on every dictation with nobody
+    /// vocabulary pass reach the same models on every dictation with nobody
     /// asking — see `Pipeline.runPrompt` — and they keep declining in silence.
     private func askForKeyThenRetry(_ error: Error, retry: () -> Void) -> Bool {
         let spec: ModelSpec
--- a/Sources/ParrotFlow/Config.swift
+++ b/Sources/ParrotFlow/Config.swift
@@ -181,7 +181,7 @@ struct Config: Decodable, Equatable {
         var acoustic: Bool = false
 
         /// How far a decoded word's spelling may sit from a term and still be
-        /// worth a line on the judge's menu. FluidAudio's similarity, where
+        /// worth a line among the readings. FluidAudio's similarity, where
         /// 1.0 is the term written out exactly.
         ///
         /// One of the two numbers that replaced the single floor (F1). That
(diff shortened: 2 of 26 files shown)


znat:
One caveat on a piece of evidence in the commit message, raised by the session that shipped #289.

The message cites "a day of the dev log carries no line from it" as evidence the stage was already dead. That was measured **before** this change, against a binary that still wrote `vocabulary judge:`. After this lands, five `Log.write` lines open `vocabulary pass:` instead, so re-running that count will return zero for a second and entirely different reason.

Anyone re-checking whether the stage is dead should not reuse the number. The evidence that stands on its own is structural and still checkable: there is no model call in the stage, and `SentenceGate` says so in its own comment — *"there is no model, and handing a place back with nobody to hand it to is a keep."*

PR #294, opened 2026-09-08
fix: a grammar edit is not a vocabulary rule

znat:
Three things the offer after a hand edit got wrong, all found in one session of using the app. Correcting `passez` to `passé` was offered as a vocabulary rule, which would spell one tense as the other in every French sentence afterwards. Typing `(___)` after `mask` was offered too, though nothing was replaced. And a correction in a field holding several dictations drew almost the whole field on the pill.

Recall does not move: 88% before and after, on the 145 corrections recorded in the dev trace. Precision goes from 44% to 46%.

Start the review at `EditWatch.inflected` and `EditWatch.added`. Both are pure functions with cases in `tests/edit-diff-cases.yaml`, so `scripts/check-edit-diff.sh` scores what ships rather than a copy of it.

<details>
<summary>Sound cannot see inflection, and the lemma can — tense, gender and number in one test</summary>

Two forms of one word are homophones. `passez` and `passé` both come out /pase/, so `soundsAlike` scores them 1.00 and overrules the word lists. That override is right for `borderplay` -> `boilerplate` and wrong here, and no sound test can separate the two.

`NLTagger`'s lemma can: both forms return `passer`. It is the dictionary form, so one test covers every kind of agreement rather than a list of endings per language:

|  | pair | lemma |
|---|---|---|
| tense | `passez` / `passé`, `run` / `ran`, `walk` / `walked` | passer, run, walk |
| gender | `grand` / `grande`, `le` / `la`, `heureux` / `heureuse` | grand, le, heureux |
| number | `cheval` / `chevaux`, `user` / `users`, `mouse` / `mice` | cheval, user, mouse |

Both sides must have a lemma, which is what keeps the test away from a name. `Prezi`, `borderplay`, `Mik` and `lectour` have none at all, so it can only ever fire on two known words. Ten controls were checked and all stay offered: `Erik`/`Eric`, `Mick`/`Mik`, `Prezi`/`Praisy`, `Ghost`/`Ghostty`, `borderplay`/`boilerplate`, `pass`/`pause`, `when`/`Qwen`, `John`/`Jon`, `Spacey`/`Spacy`, `lectour`/`Lectoure`.

Asked lowercased, the same reason `Vocabulary.unseenWord` does it: `Passez` and `Passé` at the head of a sentence come back as two nouns lemmatised to themselves.

Asked before the sound rule rather than inside it. Both refuse, but the word lists get there first whenever the two forms do not sound alike — `grand` -> `grande` — and "ordinary English" is the vaguer of two true answers.

</details>

<details>
<summary>Measurements — 63 offers to 61, no mishearing lost</summary>

`scripts/score-offers.py` against a binary built from `origin/main`, same trace, same 139 proposals:

| | offers | real | noise | precision | recall |
|---|---|---|---|---|---|
| before | 63 | 28 | 35 | 44% | 88% |
| after | 61 | 28 | 33 | 46% | 88% |

The two it drops are `passez` -> `passé` and `user` -> `users`. Neither is a mishearing by the hand label, so recall does not move.

On the 145 corrections in the trace as it now stands, the two rules catch:

    another form   passez -> passé   user -> users   page -> pages   show -> showed
    added beside   mask -> mask (___)   guitar -> "guitar "

</details>

<details>
<summary>What it does not reach — a pair macOS holds as two words</summary>

`chien` / `chienne` and `acteur` / `actrice` lemmatise to themselves rather than to a shared form, so they are still offered. They are separate lexemes to `NLTagger`, and nothing here overrides that.

</details>

<details>
<summary>Nothing was replaced when a mark is typed beside a word, and `added` could not see it</summary>

`added` is already the test for this — it drops `morning` -> `morning Tasmeen`. It missed `mask` -> `mask (___)` because it strips trailing marks off both readings before comparing, and here the whole addition is marks. `mask.` and `mask (___).` both collapse to `mask`, the two are equal, and it returns early. The evidence it needed was in what the strip threw away.

It now reads the raw difference when the stripped readings match, and the space in front of what was typed is what says it went beside the word:

    mask.    -> mask (___).    " (___)"   nothing was replaced
    guitar   -> "guitar "      " "        nothing was replaced
    small    -> small.         "."        the word changed
    Praisy   -> Praisy's       "'s"       the word changed
    rebase.  -> rebase:        "."        the word changed

The last three are pinned cases that had to keep working. The change is dropped rather than refused, which is what `added` already does, so there is no offer and no `kind: edit` row.

`inflected` no longer takes a lemma from a reading with whitespace in it. `Tagger` omits punctuation, so `mask (___)` came back as the single token `mask` and was being refused as `mask` in another form — the right answer for the wrong reason.

</details>

<details>
<summary>The pill drew four sentences because it counted words, and a run-together field has none</summary>

`learnPayload` keeps five words either side of the change. That assumes a stop has a space after it. A field holding several dictations has none: `sports. John is a painter.Jon is a musician.` is seven words holding three sentences, so five either side is four sentences of screen.

The window now cuts at the stop itself, before the words are counted. A stop counts when a letter comes before it and a space, the end, or a capital comes after, so `painter.Jon` is a boundary and `3.5` is not. On the reported case the pill goes from 508px to 153px:

    before  … basketball team.Erik likes individual sports. John Jon is a painter.Jon is a …
    after   John Jon is a painter.

The stop that ends the change's own sentence is on the pair rather than in the words after it — `trimmed` takes it off both readings and `learnPayload` puts it back. Asked over the following words alone, `Ghost.` -> `Ghostty.` ran on into the next sentence.

Also drops the gap in front of the struck word. `before` is empty whenever the change opens its sentence, which is most of them now, and the view drew the space anyway.

</details>

<details>
<summary>Review notes — where the risk is</summary>

Eleven new cases in `tests/edit-diff-cases.yaml`. `scripts/check-edit-diff.sh` gained a `lang:` column, so a French case is tagged in French, and a `learn:` column that checks the line the pill draws.

The riskiest line is the raw-difference test in `added`, because it decides whether a change is recorded at all. The three cases that had to survive it — `small` -> `small.`, `Praisy` -> `Praisy's`, `rebase.` -> `rebase:` — are in the set.

`EditWatch.offers` now takes a `language:`. It is resolved once at the caller and read by both the ear and the tagger; two sources for one field is how they disagree.

I first fixed the `mask` case by trimming the ends in `onlyPunctuation`, then reverted it. With the `added` rule in place that symptom does not occur anywhere in the 145 recorded corrections, and an unmeasured guard is worse than none.

</details>

### DIFF
--- a/Sources/ParrotFlow/AppDelegate.swift
+++ b/Sources/ParrotFlow/AppDelegate.swift
@@ -3989,38 +3989,54 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// Words kept either side of the change on the pill.
     static let learnWindow = 5
 
-    /// A stop that ends a sentence rather than sitting inside a word.
+    /// Marks that belong to the sentence they close rather than to the next
+    /// one. A stop inside a quotation is still the end of a sentence.
+    private static let closers: Set<Character> = [
+        "\"", "'", "\u{201D}", "\u{2019}", ")", "]", "}", "\u{BB}",
+    ]
+
+    /// Where the sentence ending at `i` stops, or nil if the mark there ends
+    /// no sentence.
     ///
     /// A letter before it and a space, the end, or a capital after it.
     /// `painter.Jon` is two sentences a dictation ran together and `3.5` is
     /// one number; the character alone cannot tell them apart, and the pill
     /// showing half a number is worse than the pill showing one word too many.
-    private static func endsSentence(_ text: [Character], at i: Int) -> Bool {
-        guard ".?!".contains(text[i]), i > 0, text[i - 1].isLetter else { return false }
-        guard i + 1 < text.count else { return true }
-        let next = text[i + 1]
-        return next.isWhitespace || next.isUppercase
+    ///
+    /// The index returned is past any closing quote or bracket, so `he said
+    /// "hello." Jon` cuts after the quotation and not before it. Read as the
+    /// next character, `\u{201D}` is neither a space nor a capital and the whole
+    /// sentence was kept.
+    private static func sentenceEnd(_ text: [Character], at i: Int) -> Int? {
+        guard ".?!".contains(text[i]), i > 0, text[i - 1].isLetter else { return nil }
+        var next = i + 1
+        while next < text.count, closers.contains(text[next]) { next += 1 }
+        guard next < text.count else { return next }
+        guard text[next].isWhitespace || text[next].isUppercase else { return nil }
+        return next
     }
 
     /// Whether a sentence ends anywhere in `text`.
     static func closesSentence(_ text: String) -> Bool {
         let chars = Array(text)
-        return chars.indices.contains { endsSentence(chars, at: $0) }
+        return chars.indices.contains { sentenceEnd(chars, at: $0) != nil }
     }
 
     /// What is left of `text` after the last sentence that ends inside it.
     static func afterLastStop(_ text: String) -> String {
         let chars = Array(text)
         var cut = 0
-        for i in chars.indices where endsSentence(chars, at: i) { cut = i + 1 }
+        for i in chars.indices {
+            if let end = sentenceEnd(chars, at: i) { cut = end }
+        }
         return String(chars[cut...]).trimmingCharacters(in: .whitespaces)
     }
 
     /// `text` up to and including the first stop that ends a sentence.
     static func toFirstStop(_ text: String) -> String {
         let chars = Array(text)
-        for i in chars.indices where endsSentence(chars, at: i) {
-            return String(chars[...i])
+        for i in chars.indices {
+            if let end = sentenceEnd(chars, at: i) { return String(chars[..<end]) }
         }
         return text
     }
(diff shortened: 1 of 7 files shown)


Sources/ParrotFlow/AppDelegate.swift
```diff
@@ -3984,6 +3989,58 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+    /// one. A stop inside a quotation is still the end of a sentence.
+    private static let closers: Set<Character> = [
+        "\"", "'", "\u{201D}", "\u{2019}", ")", "]", "}", "\u{BB}",
+    ]
+
+    /// Where the sentence ending at `i` stops, or nil if the mark there ends
+    /// no sentence.
+    ///
+    /// A letter before it and a space, the end, or a capital after it.
+    /// `painter.Jon` is two sentences a dictation ran together and `3.5` is
```
znat, line 4002: Fixed in 12ae32d4 — confirmed against the binary first, and you were right on the lead side.

    before  He said “hello.” John Jon is a painter.
    after   John Jon is a painter.

`sentenceEnd` now steps past any closing quote or bracket before it reads the next character, and returns the index past the closer, so the cut falls after the quotation rather than before it. Two cases added to `tests/edit-diff-cases.yaml`, one for `”` and one for `)`.

The tail side was already correct: `toFirstStop` found the stop that ends the sentence holding the change, so it did not run on. `3.5` is unaffected — the test still requires a letter in front of the stop.

PR #296, opened 2026-09-09
feat: a dictated date or time is written the shape it was said

znat:
A dictated date already carries its format. "le dix du douze" is `10/12` because the speaker chose day-then-month; "March third twenty twenty-six" is `March 3, 2026` because they chose month-then-day. This adds `examples/transforms/dates/`, which writes that shape. No model, no network, standard library only.

**A language is a file.** `en.py` and `fr.py` are two transforms, each self-contained: its own word tables, month names, cue words, readers and rules. `engine.py` beside them holds the apply loop, the JSON protocol and the `--when` printer, and knows no rule, no word and no output format. **English is the step that ships.** French is written and scored, and its pipeline step sits in `config.example.yaml` as a commented pair of lines with its gate on it, so turning it on is deleting two `#`. Adding Spanish is copying `fr.py` to `es.py`, editing the tables, and adding two lines to `config.yaml`. The steps are listed in `engine.py`'s docstring.

It reads no calendar: "next Tuesday" is not resolved and a date said without a year does not get one. It fires only on a cue — `at`/`à`/`vers`, am/pm, o'clock, past/to, `heures`, `midi`, `et demie`, a month name — because "three thirty" on its own is "three thirty-year-olds" as often as it is a time. 37 of the 77 cases are text that looks like a date and is not.

This does not depend on the numbers PR any more. Both touch `config.example.yaml` in the same place, so expect a small merge there: my `- transform: dates_en` line sits directly above where `- numbers` is today. The `Makefile` `SCORERS` line and the `checks.yml` scorer step are the same shape on both branches, so that hunk conflicts on one line.

A reviewer should start with the judgement calls in the two docstrings. Each one has a case in the set, and each is a place where a different answer is defensible.

<details>
<summary>Why it runs before `numbers`, measured on the numbers stage</summary>

Both stages read the same words, and only one can go first. Through `numbers` alone:

| said | numbers writes |
|---|---|
| "three thirty" | three thirty |
| "at ten fifteen" | at ten fifteen |
| "half past two" | half past two |
| "March third" | March third |
| "trois heures et demie" | trois heures et demie |
| "le trois décembre" | le trois décembre |
| "à dix heures quinze" | à 10 heures 15 |
| "le dix du douze" | le 10 du 12 |

The last two are the reason. Once `numbers` has run, the words a date is made of are gone. So the dates steps read them first. They also accept digits, because the decoder emits them either way — "March 3rd 2026", "at 10 15".

Verified in the other direction. `--pipeline` with `dates_en`, `dates_fr`, then `numbers`:

```
in:   the deadline is March third twenty twenty-six
  → transform
      the deadline is March 3, 2026
      dates_en.applied = "month day"  dates_en.count = 1  dates_en.language = "en"
  ⊘ transform  — skipped, when /(?i)…/ did not match
  · numbers  — ran, changed nothing
out:  the deadline is March 3, 2026

in:   on commence à dix heures quinze
  ⊘ transform  — skipped, when /(?i)…/ did not match
  → transform
      on commence à 10h15
      dates_fr.applied = "heures"  dates_fr.count = 1  dates_fr.language = "fr"
  · numbers  — ran, changed nothing
out:  on commence à 10h15

in:   we took two fifteen minute breaks
  ⊘ transform  — skipped, when /(?i)…/ did not match
  ⊘ transform  — skipped, when /(?i)…/ did not match
  · numbers  — ran, changed nothing
out:  we took two fifteen minute breaks
```

`numbers` leaves every output of these stages alone, and each language's gate skips the other language's sentence.

</details>

<details>
<summary>51/51 English, 26/26 French, and 3 correct edits to 925 archived clips</summary>

`ParrotFlow --eval dates_en` and `--eval dates_fr`, against a config built from `config.example.yaml` in a scratch directory. `tests:` on each transform is what sends `--eval` to the right case file, so neither needs `--cases`.

```
eval: dates_en                          eval: dates_fr
  cases    cases-en.yaml  (51)            cases    cases-fr.yaml  (26)
  change   25/25 = 100%                   change   15/15 = 100%
  keep     26/26 = 100%                   keep     11/11 = 100%
  overall  51/51 = 100%                   overall  26/26 = 100%

  by probe                                by probe
    date_en           10/10                 date_fr        5/5
    time_en           9/9                   time_fr        10/10
    time_en_fraction  6/6                   not_a_date_fr  3/3
    not_a_date_en     10/10                 not_a_time_fr  8/8
    not_a_time_en     16/16

  latency  0.05s median, 0.05s worst     latency  0.05s median, 0.06s worst
```

`score.py` gives the same numbers without building Swift, per language, and adds two checks `--eval` cannot make: every `change` case must match that language's `when:` regex, and the `when:` lines in `config.example.yaml` must be the ones `--when` prints.

```
  en   51/51   change 25/25   keep 26/26   when: 25/25 change and 10/26 keep through the gate
  fr   26/26   change 15/15   keep 11/11   when: 15/15 change and 8/11 keep through the gate
```

A keep case the gate rejects is safe — the script never runs on it — but it tests the script rather than the pipeline. That is 16 English and 3 French cases.

**Latency.** 0.05 s median warm per step, against `repetitions` at 0.03 s on the same binary. A one-shot `--pipeline` run reports about 385 ms; that is the pyenv shim probe, paid once per process, not the stage — see `docs/architecture.md`, "The version-manager shim".

**The archive.** `score.py --corpus` over 925 clips of real dictation. The English gate opens on 24, the French on 3, and 3 clips are edited:

```
  -  Two, could remove our one on one to one pm? Thanks, Nathan.
  +  Two, could remove our one on one to 1 pm? Thanks, Nathan.

  -  Are you available at eleven AM?
  +  Are you available at 11 AM?

  -  ... has been merged since the last one on July third.
  +  ... has been merged since the last one on July 3.
```

All three are right. A fourth — "since July 3rd" to "since July 3" — was wrong and is now a keep case and a guard.

</details>

<details>
<summary>Every judgement call, and where a different answer would be defensible</summary>

English, in `en.py`:

- **"seven o'clock" stays `7 o'clock`**, not `7:00`. The speaker said no minutes and `:00` is two digits they did not dictate.
- **"noon" and "midnight" stay words.**
- **Minutes under ten need "oh five" or "05".** "at two three" is not 2:03 in anybody's mouth.
- **A bare `H MM` caps the hour at 12.** "at twenty twenty-six" is a year. The cost: "at fifteen thirty" is left alone.
- **`MM to H` needs a lead cue.** "five to ten people" and "from ten to twelve" are ranges. `MM past H` does not: "past" followed by an hour is not a range.
- **Bare "am" needs a lead cue.** "one am ready" is a sentence.
- **"3pm" and "July 3rd" are left alone.** The digits are there and the only edit left is a space or a suffix — typography, not a date being written down.
- **A lone ordinal and a lone year are left alone.** "the fifteenth" and "twenty twenty-six" are the numbers stage's job.
- **"second" is an ordinal here.** The numbers transform drops it so "a thirty second timeout" survives. Every ordinal here stands next to a month name or another ordinal, where a duration cannot be meant.

French, in `fr.py`:

- **A bare `H heures` fires at 13 and above, or after à/vers.** "il y a trois heures" is a duration; "vingt heures" is a time. Known miss: "j'ai dormi vingt heures" becomes "j'ai dormi 20h".
- **"a" is not "à".** The unaccented verb was in the lead list and turned "il a trois heures d'avance" into "il a 3h d'avance". Accented forms only, and an elided `d'` after the hour declines it.
- **"midi" and "minuit" stay words on their own**; "midi et demi" becomes `12h30`, because there the half hour is dictated.
- **"le" stays.** "le 3 décembre", "le 10/12". English "the third of March" drops its "the", because "the 3 March" is not English.
- **"premier" is written `1er`.**
- **Years are read up to "deux mille soixante-neuf".** Above that, digits.

Shared, in `engine.py`:

- **No step declines on `ctx.language`.** It is read and published as `<stage>.language`, and never acted on. A French clock phrase inside an English transcript is still a French clock phrase, and the cue words barely overlap, so the wrong script writes nothing. Declining would only cost the mixed sentences.

</details>

<details>
<summary>Why the word tables are duplicated instead of shared</summary>

An earlier draft imported the word tables from the numbers transform. It read well and it was wrong for this folder: it made `dates/` unable to stand on its own, it put a second package's release schedule in front of adding a language, and "copy a file and edit it" stopped being the whole job.

Each language file now carries about forty words of its own. They do not move. One shared table is one place every language can break, and the cost of the copy is paid once per language, by the person adding it, in the file they are already editing.

Nothing outside `engine.py` is shared, and `engine.py` has no words in it.

</details>

<details>
<summary>A finding, not fixed here: `--eval` dies with SIGPIPE on a case file that carries its own command transform</summary>

A case file can carry a `transforms:` section so it states the transform it assumes instead of inheriting the machine's config. With a `command:` whose path is relative, `--eval` exits 141 and prints nothing at all — not even its header.

Reproduced with a two-line script, nothing to do with this change:

```yaml
# /tmp/mini/cases.yaml
transform: mini
transforms:
  - name: mini
    command: mini.py          # exit 141, no output
    returns: json
cases:
  - input: "hello"
    expect: "hello"
```

`command: /tmp/mini/mini.py` works. `command: ./mini.py` works. A `replace:` body works. The bare name resolves in `transforms/<name>/` under the file's own directory and nowhere else — `TransformFolder` says so deliberately — but an unresolved command should be reported, not fatal.

So neither case file carries a `transforms:` block, and `--eval` reads the transforms from the config. Say the word and I will file it.

</details>

<details>
<summary>Review notes</summary>

- Start with `engine.py`. It is 150 lines and it defines what a language file has to expose: `RULES`, an ordered list of `(name, compiled regex, handler)`, and `gate_parts()`. A handler returns the rewrite or `None`.
- Then `fr.py`, which is the shorter of the two languages and the one you would copy.
- `cases-en.yaml` and `cases-fr.yaml` are the same 77 cases as before, split by `lang`. The names match the numbers transform on #298.
- `config.example.yaml` gets one pipeline step above `numbers`, one commented step under it for French, and two `transforms:` entries. The `numbers` line itself is untouched — it changes on the other branch.
- `score.py` runs in CI as its own step, `Dates transform`, and in `make test` through a new `SCORERS` list. `scripts/check-dates.sh` stays: it scores `DateRewriter.swift`, which this branch does not touch.
- The `when:` lines are generated. `<lang>.py --when` prints one; `engine.alt()` breaks length ties alphabetically so it prints the same line every run, and `score.py` fails if the config has drifted. It reads the commented French line the same way, because a comment is still text in the file.
- Docs: one row in `README.md`, one in the `docs/authoring.md` table, one sentence in `docs/pipelines.md`, and one sentence in the `DateRewriter.swift` header pointing at the transforms. `DateRewriter.swift`, `DatesCommand.swift` and `--dates` are otherwise untouched.
- Checks run green: `check-default-config.sh`, `check-transform-folders.sh`, `check-pipeline-config.sh`, `check-dates.sh`, `check-eval.sh`, and `swift build -c release`.

</details>


znat:
Replaced by #303, which merges this and #298 into one branch off main. Same dates transform, plus: no language gate on the steps, French shipped as files with the config naming English only, and the file headers trimmed. Scores are unchanged — dates en 53/53, fr 80/80.

PR #298, opened 2026-09-09
feat: spoken numbers are a Python transform, and they can write a percent sign

znat:
`numbers` was a built-in Swift stage. It is two shipped Python transforms now, `numbers_en` and `numbers_fr`, one runnable script per language over a shared `engine.py`. They do the same job on the same cases, and they can write a percent sign: "seventy-five percent" and "soixante-quinze pour cent" both become `75%`. 1,050 lines of Swift left the app.

**This runs a program on every dictation that contains a number word.** Both are `command:` transforms, like `punctuation` and `repetitions`, and `config.example.yaml` declares them. Only the English one is a pipeline step by default; the French step is there commented out. Each step carries a `when:` gate printed by its own script, so the python3 start is paid only when the transcript could hold a number of that grammar.

**A config that still says `- numbers` is refused, not ignored.** `--check-config` names the stage and spells out both transforms and both steps — that config was converting French numbers, so the advice restores it. Half a migration is refused too.

Review first: the split into `engine.py` + `en.py` + `fr.py`, then the cross-language guard, then the `when:` gates on the pipeline steps.

<details>
<summary>A language is a file, so adding one is a copy</summary>

```
engine.py   the tokeniser, the parser, percent, the guard, the IO
en.py       the English tables and three rules, and nothing else
fr.py       the French tables and three rules, and nothing else
score.py    scores every language it finds beside it
```

`engine.py` names no language and is not executable. Adding one is `cp fr.py es.py`, edit the tables, copy the case file, add a `transforms:` entry and a pipeline step. Duplication between the language files is accepted on purpose: the alternative is one file that every language has to be read through.

`score.py` and the CI step find the languages by looking, so a third pair of files is scored without editing either.

</details>

<details>
<summary>The port is faithful: 3,194 lines compared, 0 disagreements (measured before the split)</summary>

This comparison was run against the first version, one script holding both grammars, before it became `engine.py` + `en.py` + `fr.py`. The split is held to the 99 case-file cases and the 19 pipeline cases below instead.

Every judgement in `Numbers.swift` moved across — where a number begins and ends, digit runs, years, ordinals, decimals, the below-ten floor for a lone word, the vigesimal rule, septante/huitante/nonante, and the cross-language rule with its "99 cents" guard.

Before deleting the Swift, both were run over the same 3,194 lines with one grammar pinned: the 97 case inputs under both grammars, plus 3,000 generated permutations of number words, filler words, hyphens and commas.

```
3168/3194 agree, 26 differ only by the new percent rule, 0 disagree
```

The 26 are all percent. After the "no percent on an ordinal" guard was added, the remaining differences are exactly the intended new behaviour.

</details>

<details>
<summary>Percent: 75% in both languages, and "pour cent" with no number in front stays</summary>

| Said | Written |
|---|---|
| seventy-five percent | `75%` |
| two per cent of them | `2% of them` |
| a hundred percent | `100%` |
| five percent | `5%` |
| soixante-quinze pour cent | `75%` |
| cent pour cent | `100%` |
| trois virgule cinq pour cent | `3,5%` |

Four decisions:

- **No space, in French too.** French typography wants a narrow no-break space. This writes `75%` in both languages, deliberately: the transcript is pasted into terminals and code fields as often as into prose. Said in the script's docstring.
- **Percent lifts the below-ten floor.** Everywhere else a lone number word under ten stays a word. `five%` is never right, so "five percent" is `5%`.
- **The other direction is left alone.** "pour cent" with no number in front is the preposition and a hundred. "il paie pour cent euros" is untouched, and so is "percent of the time is what matters". That rule predates percent — it is what kept "soixante-quinze pour cent" from becoming "75 pour 100" — and the set has cases for both directions.
- **Never on an ordinal.** "the fifth percent of the sample" is not a percentage, and `5th%` is not a thing anyone types.

The one existing case that flips: `soixante-quinze pour cent` was `75 pour cent` and is now `75%`.

</details>

<details>
<summary>Scores: en 38/38, fr 61/61, and 19 pipeline cases for what neither file can score</summary>

Each language has its own case file, `cases-en.yaml` and `cases-fr.yaml`, and each case pins `lang:`, so what is scored is the grammar.

```
$ examples/transforms/numbers/score.py

  en  38/38   change 27/27   keep 11/11
      cardinals 10/10   extras 6/6   keep 10/10   percent 12/12

  fr  61/61   change 49/49   keep 12/12
      belgium 9/9   extras 5/5   floor 3/3   keep 10/10
      percent 9/9   scales 9/9   vigesimal 16/16

  99/99 over 2 language(s)
```

`--eval` scores the same sets through the real pipeline step, against the installed copy:

```
$ ParrotFlow --eval numbers_en     overall 38/38 = 100%   0.03s median
$ ParrotFlow --eval numbers_fr     overall 61/61 = 100%   0.03s median
```

Three things belong to no single file — language detection, the cross-language guard, and the order of the two steps. They were the old `lang: auto` cases. They are `--pipeline` cases now, against `tests/pipelines/numbers.yaml`, and `scripts/check-pipeline.sh` is 113/113 with them (94/94 before).

`check-pipeline-config.sh` is 85/85, `check-dotted.sh` 81/81, `check-transform-folders.sh` 30/30, `check-eval.sh` 29/29, `check-default-config.sh` green.

</details>

<details>
<summary>The cross-language guard is asked about the number, not about the sentence</summary>

This matters once the French step is on. A step is not gated on `language ==`: the language of a sentence does not decide which numbers are in it. "on a mergé la pull request avec vingt et un commits" is French with English in it, and English sentences carry French numbers the other way round.

That needs a guard, or French reads the `cents` in "I have 99 cents" as hundreds. Above four words, a script whose language is not `ctx.language` only writes a number whose **own words** include a unit, a teen or a tens word of its grammar. A bare scale word is not enough, and an unrelated number word elsewhere in the sentence cannot vouch for it.

```
I have 99 cents                       unchanged
it costs 50 cents each                unchanged
we made two hundred cents on six      we made 200 cents on six
j'ai payé deux cents euros hier       j'ai payé 200 euros hier
j'en ai quatre cent septante-huit     j'en ai 478
```

Below four words `NLLanguageRecognizer` cannot answer, so every grammar tries — otherwise "cent euros" and "vingt et un" arrive as English and come back untouched.

`numbers.language` is gone. Each script publishes its own `numbers_en.count` / `numbers_fr.count`, and the `languages` scope line added in the first version is gone with it, since nothing else read it.

</details>

<details>
<summary>The `when:` gates are generated, and score.py fails if a case misses one</summary>

`en.py --when` and `fr.py --when` each print a case-insensitive regex built from that script's own word tables — every number word of that grammar, cardinal and ordinal. Connectors ("and", "point", "et", "virgule") are left out: none of them makes a number without one of those words beside it.

Those regexes are what `config.example.yaml` carries: on the English step, and on the commented-out French one. `score.py` reads each regex **from the script, not from the config**, and fails if any case that must change does not match it. A gate that misses a word is a silent miss: the step does not run, the number stays a word, and nothing shows it happened.

Checked through `NSRegularExpression`, which is what compiles a `when:` pattern, not just through Python's `re`.

| Input | Ran | Out |
|---|---|---|
| le quinzième essai | `numbers_fr` | le 15e essai |
| ZÉRO VIRGULE CINQ | `numbers_fr` | 0,5 |
| j'en ai quatre cent septante-huit | `numbers_fr` | j'en ai 478 |
| soixante-quinze pour cent | `numbers_fr` | 75% |
| on se voit demain | neither | unchanged |

The gates in `config.example.yaml` and in `tests/pipelines/numbers.yaml` were diffed against what the scripts print. Both match, character for character. Nothing checks that automatically yet: `score.py` reads the script, never the config, so a word table edited without regenerating the config gate is still a silent miss.

**Only the English step is on by default.** The French gate has to contain `un` and `une` — "un pour cent" is a number — so it opened on most French sentences and paid a python3 start there for everyone, including people who never dictate French. On 67 French lines taken from `tests/` it fires on 21; without `un`/`une`, on 3.

So `config.example.yaml` lists `- transform: numbers_en` and carries the French step under it commented out, gate and all:

```yaml
    - transform: numbers_en
      when: /(?i)\b(?:seventeenth|…)\b/
    # French too, if you dictate in it. The gate is `fr.py --when`.
    #- transform: numbers_fr
    #  when: /(?i)\b(?:cinquantieme|…)\b/
```

The `numbers_fr` transform stays defined, `score.py` still scores both languages, and CI still runs it. Turning French on is uncommenting two lines. `tests/pipelines/numbers.yaml` keeps both steps: detection, the cross-language guard and the order of the two scripts only exist when both are on.

</details>

<details>
<summary>Migration: three shapes refused, each with what to write</summary>

```
$ ParrotFlow --check-config      # pipeline: [vocabulary, numbers]
  ✗ pipeline: "numbers" is two shipped transforms now, one per language, not a
    built-in stage. Add both to `transforms:` — `name: numbers_en`,
    `command: examples/numbers/en.py`, `returns: json`, and the same with `_fr`
    and `fr.py` — then write `- transform: numbers_en` and
    `- transform: numbers_fr` in the pipeline, in that order. Each step takes
    the `when:` gate its own script prints: `examples/numbers/en.py --when`
```

The same sentence answers a `- transform: numbers` that names nothing, a `- transform: numbers_en` whose `transforms:` entry is missing, and `transcription.numbers: true`. All exit 1. The first is the one every existing config hits.

</details>

<details>
<summary>Latency: 2 ms of Swift became about 55 ms per step, paid only when a gate matches</summary>

Measured on "we spent two hundred and forty three dollars, seventy five percent of it", six runs each.

| | ms |
|---|---|
| The Swift stage (shipped build) | 2 ms, every run |
| One step, `/opt/homebrew/bin/python3` | 37 ms, 5 runs |
| One step, `/usr/bin/python3` | 48 ms, 5 runs |
| Under `--eval`, warm | 0.03 s median |
| The first version, one script (`--pipeline --vars`) | 51–60 ms, median 55 |

The difference is the python3 start, which `CommandRunner` documents at 40–60 ms. An English sentence with no French number word pays one start, not two: the French gate does not match. 1,050 lines of Swift left the app.

**One warning about measuring this.** `--pipeline … --vars` on this machine reports `numbers_en.ms` at 274–325. That is not the script. `#!/usr/bin/env python3` resolves to `~/.pyenv/shims/python3` here, and the shim adds about 215 ms per start. It hits every `command:` transform equally and is not new — the same script under a real interpreter is 37 ms. Measure with the interpreter named, or the number describes the shim.

`docs/architecture.md` had one row reading "`the vocabulary stage`, `fuzzy`, `numbers`, a `replace:` transform | 0.035 s". That row covered a group, not the numbers stage alone. `numbers` is out of it now, and the row below it already gives the cost of a `command:` transform.

</details>

<details>
<summary>What moved in the tree, and the one thing worth arguing about</summary>

Deleted: `Numbers.swift`, `NumberGrammar.swift`, `NumberGrammar+English.swift`, `NumberGrammar+French.swift`, `NumbersCommand.swift`, the `--numbers` flag, `scripts/check-numbers.sh`, `tests/numbers-cases.yaml`, and `case numbers` in `Pipeline.Stage`.

Added: `examples/transforms/numbers/{engine.py,en.py,fr.py,cases-en.yaml,cases-fr.yaml,score.py}` and `tests/pipelines/numbers.yaml`.

**`Pipeline.everything` no longer contains numbers.** It is derived from `Stage.allCases`, so a config with no `pipeline:` key at all now runs `interpret` and nothing else. That was already true of `vocabulary` and of every transform. `config.example.yaml` lists the English step, and a config that has a `pipeline:` block with `- numbers` in it is refused rather than silently dropped, so the only person affected is one who deleted the whole block. Not given a `notices()` line; say so if you want one.

The `--pipeline` fixtures use symlinks, `tests/pipelines/transforms/numbers_en/en.py` and `numbers_fr/fr.py`, the same pattern `code_identifiers` already uses there.

</details>

<details>
<summary>One pre-existing bug this makes easier to hit</summary>

A `returns: json` transform whose `command:` cannot be resolved kills `--pipeline` with SIGPIPE instead of failing open. Reproduced on a fixture that has nothing to do with numbers:

```yaml
transforms:
  - name: ghost
    description: names a script that is not there
    command: ghost.py
    returns: json
pipeline:
  - transform: ghost
```

Without `returns: json` the same fixture fails open correctly and prints the text unchanged. This is not new here — `punctuation` has the same shape — but numbers adds two more `returns: json` commands, so it is worth its own issue.

</details>


znat:
Replaced by #303, which merges this and #296 into one branch off main. Same numbers transform, plus: no language gate on the steps, and a bare scale word from a million up is no longer read as a number, so "des millions de personnes" and "three million dollars" come back untouched. Scores: en 38/38, fr 63/63.

PR #321, opened 2026-09-17
refactor: the pill drops the floating capsule it no longer wears

znat:
The pill has hung off the line of text since `isDocked` was pinned to `true`, so
the undocked capsule it used to wear was unreachable: `Dock` was never nil, and
the only thing still drawing that style was four models on `--panel-sheet`. This
removes the style, the `Dock?` that carried it, and the parts that existed only
for it. Nothing the app draws changes, with one exception: the microphone
permission screen sized its example pill from the capsule's width, so the tab it
really draws was stretched to 114x42 instead of its own 38x27 — it asks
`panelSize` now. Start the review at `PillHUD.beside` and `PillView.shape`,
where the nil case is gone, and at `build()`, which no longer wraps the hosting
view in glass.

`checks` is green, so it compiles and SwiftLint is clean. It was written without
a build: this container has no Swift toolchain, so each removed symbol was
checked by reading its call sites and CI was the first compile. The panels are
looked at rather than asserted on, so `--panel-sheet` on a screen is still owed —
the surfaces to compare are the permission screen's example pill and the four
rows the sheet lost.

<details>
<summary>Why the style was unreachable — one line in PillHUD said so</summary>

`PillHUD.isDocked` was `private var isDocked: Bool { true }`, with a comment
saying "Always, now". Every path that could answer nil went through it:

- `anchor(_:)` returned `isDocked ? .free : nil`
- `beside(...)` returned `isDocked ? dock : nil`
- `wantedDock` returned `.free` or `.below`, never nil

So `PillModel.docked` could only ever be `.below`, `.above` or `.free`, and
every `dock == nil` branch under it was dead: the capsule in `RecordingContent`,
the sentence in `MessageContent` for the `.working` state, the lozenge in
`PillView.shape`, and the widths and heights in `PillMetrics`.

</details>

<details>
<summary>Deleted, by symbol — 323 lines out, 99 in</summary>

| Gone | Was |
|---|---|
| `PillHUD.isDocked`, `PillView.isDocked` | both answered the same question twice |
| `PillHUD.dock(_:)` | set the dock and moved the glass; now one assignment |
| `PillHUD.backdrop`, `backdropIsIn` | the glass under the capsule |
| `Meter` | the twelve bars |
| `PillMetrics.floatGap`, `recording(hasIcon:label:)`, `meter`, `icon`, `tuck` | measured the capsule |
| `PillView.floating` | the capsule's corner radius |
| `Dock?` in `PillModel.docked`, `anchor`, `beside`, `panelSize`, `width`, `height` | now `Dock` |
| four `pill(...)` models on the sheet | `floating-recording`, the same with no icon, `floating-command`, the floating `working` pill |

`MessageContent` drew two surfaces off one `dock == nil` test. It is split:
`WorkingContent` is the bird tab, `MessageContent` is the notice's sentence.

</details>

<details>
<summary>Dropping the glass changes nothing on screen — it was taken out on every placement</summary>

`build()` wrapped the hosting view in `ParrotGlass.container` and kept the
backdrop in a property. `dock(_:)` put it back only when the dock was nil, and
took it out otherwise. Since the dock was never nil, the first placement removed
it and nothing ever put it back — and that placement runs before `fadeIn`, so it
never reached the screen. The panel now holds the hosting view directly.

`ParrotGlass.container` is untouched. `LaunchPanel` and `PreviewPanel` still use
it.

</details>

<details>

Two leftovers of the same move, both older than this branch and both left alone:

- The permission screen's mock has no `appIcon`, so `PlumageMeter` draws it
  `blind` — the bird that means "nowhere to type". On the screen that asks for
  the microphone that says the wrong thing. `TourPill` solves it by setting an
  icon that is never drawn.
- `PillState.working(String)` still carries a message nothing draws. Removing it
  touches `AppDelegate`, `TutorialPanel`, `TutorialSlackPanel`, `ChatStage` and
  `PanelsCommand`.

</details>

- [x] `make test` passes — as far as CI runs it: the build, SwiftLint and the
  sets that need no model and no screen. The four that need Ollama or a real
  screen were not run here.
- [x] No prompt, table or pattern changed, so there is nothing to score.


znat:
@coderabbitai review

The last review covered `4458d9c`. `7452af2` adds the Combine import and removes two blank lines.

---

